Issue #17, opened 2025-01-24
chore: Make router a python package


gaocegege:
Making the router a Python package offers several benefits:

- The package can be installed with pip, simplifying dependency management and deployment.
- Tests can be run more easily, and the package can be integrated into CI/CD pipelines for automated testing.
- The package can be versioned, allowing for controlled updates and backward compatibility.

Besides, we should include some basic end-to-end (e2e) tests as part of the continuous integration (CI) pipeline.

gaocegege:
Hey, I saw the roadmap mentions a "more performant router using non-Python languages." If we plan to tackle that soon, maybe we could explore rewriting the router in another language for a performance boost and ignore this. Just a thought.

ApostaC:
Hey @gaocegege , making it a python package sounds great to me!

We are right now working on end-to-end tests and CI pipeline. And it should be available soon. Thanks for pointing that out!

gaocegege:
Thanks for the prompt response.

Should we turn it into a Python package now, or wait until the CI setup is ready? It may impact the CI implementation.

ApostaC:
It would be great if we could turn it into a python package now. (otherwise we need to adjust the CI again IIUC?)

PR #18, opened 2025-01-24
feat: Support dynamic node changes in session router

### LINKED ISSUE #8: feat: Support Dynamic Addition/Removal of Instances in Session-Based Routing
The current setup uses a basic hashing approach for routing, which doesn’t handle node changes very well. 

```py
            hash_digest = hashlib.sha256(session_id.encode()).hexdigest()
            index = int(hash_digest, 16) % len(endpoints)
            url = endpoints[index].url
            self.key_to_server_id[session_id] = URL
```

Consistent hashing is a better fit because it minimizes data shuffling when instances are added or removed, keeping things running smoothly. Maybe we could give it a try?

gaocegege:
Close #8 

All the test cases pass but it is a little tricky to run the tests:

```bash
export PYTHONPATH=$(pwd)/src/router
pytest
```

it might be a good idea to turn the router into a Python package—it could make testing a lot easier and help keep things neat and tidy.

Ref #17

### DIFF
--- a/src/router/requirements.txt
+++ b/src/router/requirements.txt
@@ -4,3 +4,4 @@ httpx
 uvicorn
 kubernetes
 prometheus_client
+uhashring
--- a/src/router/routing_logic.py
+++ b/src/router/routing_logic.py
@@ -2,7 +2,9 @@ from typing import List, Optional, Dict
 import hashlib
 import abc
 import enum
+
 from fastapi import Request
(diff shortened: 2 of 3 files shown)


gaocegege:
cc @njhill

ApostaC (approved):
Great stuff! LGTM and thanks for the PR!

PR #23, opened 2025-01-26
feat: Add Github workflows to lint and release Helm chart

0xThresh:
This PR adds Github Actions workflows that can run `helm lint` on the Helm charts, and release Helm charts to Github using Github Pages as a publishing source. 

**Action Required - Set Up `gh-pages` Branch** 
Before the Helm release workflow will work correctly, there is setup required by a maintainer. The steps can be found in the [Helm docs](https://helm.sh/docs/topics/chart_repository/#github-pages-example), and should only take about 5 minutes. 

A successful test run of this workflow can be seen in my fork here: https://github.com/0xThresh/vllm-production-stack/actions/runs/12979140685/job/36194506180 

And you can see the charts that I successfully published on my repo here: https://0xthresh.github.io/vllm-production-stack/index.yaml

While I hope this is a helpful PR, I also understand if there are other plans for chart publishing. If that is the case, I hope the Lint workflow is still helpful, and can update my PR to only include that. Thank you!

### DIFF
--- a/.github/workflows/helm-lint.yml
+++ b/.github/workflows/helm-lint.yml
@@ -1,0 +1,26 @@
+name: Lint vLLM Stack Helm Charts
+
+on:
+  pull_request:
+    paths:
+      - "helm/**"
+  push:
+    paths:
+      - "helm/**"
+
+jobs:
+  lint-chart:
+    name: Lint Helm Chart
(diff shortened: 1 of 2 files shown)


ApostaC (approved):
LGTM! Thanks for the contribution!

PR #46, opened 2025-01-30
[fix] triggers for functionality test

ApostaC:
Currently, the functionality test has not been triggered when the PR is changing the helm chart.

This PR fixes this problem


Shaoting-Feng (approved):
Thanks for the fix. LGTM.

PR #49, opened 2025-01-31
feat: Make log stats interval as a hidden CLI argument

gaocegege:
This PR is to make log stats interval configurable, it is a **hidden** argument for debug purposes. Not sure if we should make it public since it should be useless in most cases.


ApostaC (approved):
LGTM!

Issue #47, opened 2025-01-31
feat: Offline batched inference based on OpenAI offline batching API


gaocegege:
This issue is created to track the conversation about the offline batch API.

gaocegege@

> I was just wondering if there are any plans to add [openai api compatible offline batch](https://platform.openai.com/docs/api-reference/batch/create) support in the router. I saw a comment about it https://github.com/vllm-project/vllm/issues/1636#issuecomment-1816831493, and it looks like this feature needs file interfaces for uploads, which might not fit well with vLLM. It could be a nice addition to the router, acting as a bridge between users and vLLM.

ApostaC@ 

> Seems like there should be multiple design choices to handle file uploads. Will add this to the roadmap (which will be released as an issue in the project tomorrow). Maybe we can have more detailed discussion there?

simon-mo@ 

> I also agree this can be a lightweight optional component in the stack! given in K8s you can provision persistent volume easily or mount s3-fuse

gaocegege:
We could start by keeping the files in the local file system, especially since our router currently doesn't support multi-instance deployment. We can design a robust abstraction for file uploads to ensure extensibility, allowing us to support other storage backends (e.g., MinIO, S3) in the future, particularly for k8s environments.

ApostaC:
@gaocegege Thanks for leading the discussion here! 

> We could start by keeping the files in the local file system, especially since our router currently doesn't support multi-instance deployment.

The proposal looks good to me! Would you like to take a stab at designing the concrete interface in the router?

gaocegege:
Yes! I'd be happy

PR #52, opened 2025-02-01
feat: OpenAI batch API part 1

gaocegege:
This is the part 1 of #47 

file.Storage has been introduced to abstract file operations, to support both OpenAI files and the uploads API. The user-id in the interface is defaulted to uid-default, which is kept for future extensibility.

I believe OpenAI uses a database, such as PostgreSQL, to store file metadata. For now, we are using simple file operations to handle this.

I’ve raised this PR to discuss whether this approach aligns with our requirements. Then will implement the batch API in part 2.

### DIFF
--- a/src/vllm_router/files/storage.py
+++ b/src/vllm_router/files/storage.py
@@ -24,6 +24,22 @@ class Storage(ABC):
         content: bytes = None,
         purpose: str = DEFAULT_PURPOSE,
     ) -> OpenAIFile:
+        """
+        Save a file with the given parameters and returns an OpenAIFile object.
+
+        Args:
+            file_id (str, optional): Unique identifier for the file. Defaults to None.
+            user_id (str, optional): ID of the user uploading the file. Defaults to DEFAULT_USER_ID.
+            file_name (str, optional): Name of the file. Defaults to None.
+            content (bytes, optional): Binary content of the file. Defaults to None.
+            purpose (str, optional): Purpose of the file upload. Defaults to DEFAULT_PURPOSE.
+
+        Returns:
+            OpenAIFile: Object containing the saved file information.
+
+        Notes:
+            This is an abstract method that must be implemented by subclasses.
+        """
         pass
 
     @abstractmethod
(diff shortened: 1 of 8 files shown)


ApostaC:
Thanks @gaocegege . I will take a look at this PR soon!

src/vllm_router/router.py
```diff
@@ -139,6 +139,60 @@ async def route_general_request(request: Request, endpoint: str):
     )
 
 
+@app.post("/v1/files")
```
ApostaC: Just noticed the new endpoints all have this `/v1` prefix, while the existing endpoints don't have.

Do you think we should keep it consistent (maybe adding `/v1` prefix to old endpoints? we can create a issue for this)

src/vllm_router/files/storage.py
```diff
@@ -0,0 +1,59 @@
+from abc import ABC, abstractmethod
+from typing import List
+
+from vllm_router.files.files import OpenAIFile
+
+
+class Storage(ABC):
```
ApostaC, line 7: Can we add docstrings for the interfaces defined in the `Storage` class?

ApostaC (commented):
Left some comments. Otherwise LGTM!

In the meantime, can we write down the document of the new endpoints somewhere (e.g., in issue #47)?

src/vllm_router/router.py
gaocegege: I think we should remove it since all endpoints (in router and vLLM) do not have the prefix.

Users could use this:

```
client = OpenAI(
    base_url="http://localhost:8000/v1",
    api_key="token-abc123",
)
```

src/vllm_router/files/storage.py
gaocegege: Sure.

ApostaC:
@gaocegege Hey, do you want this PR to be merged? Or you want to edit it a bit more?

gaocegege:
I will address the comments first

ApostaC:
Thanks! Take your time please 👍

gaocegege:
Please take a look, thanks for your time :smile:

src/vllm_router/files/storage.py
```diff
@@ -0,0 +1,66 @@
...
+        self, file_id: str, user_id: str = DEFAULT_USER_ID
+    ) -> bytes:
+        pass
+
+    @abstractmethod
+    async def list_files(self, user_id: str = DEFAULT_USER_ID) -> List[str]:
+        pass
+
+    @abstractmethod
+    async def delete_file(self, file_id: str, user_id: str = DEFAULT_USER_ID):
```
ApostaC: Sorry for being confusing. But previously I meant we should add some docstrings for the methods. (In case in the future other contributors want to extend this class and implement other storages)

ApostaC (approved):
Just one minor comment. Other LGTM!

src/vllm_router/files/storage.py
gaocegege: SGTM!

gaocegege:
Comments are addressed, please take a look.
Thanks for your time :smile:

ApostaC (approved):
LGTM!

gaocegege:
Thanks. I will work on the part 2: the batch API.

Issue #28, opened 2025-01-28
feat: Add Ingress configuration


0xThresh:
I will need to expose models to external users through an Ingress. I plan to open a PR for this, but wanted to document this in an issue for tracking.

KuntaiDu:
Thanks for contributing 🎉

PR #54, opened 2025-02-01
feat: Add ingress configuration to routerSpec

0xThresh:
Adds an ingress configuration to the routerSpec, enabling connections to the router service from outside the cluster. 

Closes https://github.com/vllm-project/production-stack/issues/28

### DIFF
--- a/helm/templates/ingress.yaml
+++ b/helm/templates/ingress.yaml
@@ -1,0 +1,42 @@
+{{- if .Values.routerSpec.ingress.enabled -}}
+apiVersion: networking.k8s.io/v1
+kind: Ingress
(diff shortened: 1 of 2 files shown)


ApostaC:
Thanks for the contribution. Can you add a basic example of how to use the ingress?

Could be in a separate PR if you want (we can create an issue to track this).

ApostaC:
Otherwise LGTM!

ApostaC:
@0xThresh Since the issue is out there, I will merge this PR first. Thanks for contributing!

ApostaC (approved):
LGTM!

PR #65, opened 2025-02-04
fix: Update Helm values example to use new hf_token value

0xThresh:
https://github.com/vllm-project/production-stack/pull/22 updated the schema to use `hf_token` instead of an environment variable config in `Values.yaml`, but the commented out example in the values file was not updated to reflect this. This PR corrects the example to use the new hf_token value.


ApostaC (approved):
Good catch! Thanks!

PR #76, opened 2025-02-07
[Add] fix for router files api and example to post query to the api

ApostaC:
This PR fixes a problem that the router will break when the "purpose" does not exist in the query to `/files` endpoint.

It also contains a minimal example to query the `/files` endpoint.


gaocegege (approved):

gaocegege:
Thanks for the fix!

PR #83, opened 2025-02-07
feat: Wrap router to a singleton

gaocegege:
The batch implementation will also use the router, so we should implement it as a simple singleton, similar to `EngineStatsScraper` and other components.


ApostaC (approved):
LGTM!

PR #103, opened 2025-02-10
fix: set Deployment strategy as Recreate instead of RollingUpdates

amasolov:
RollingUpdates can't be used if replica count is set to 1. It would wait indefinitely to re-deploy the pods. Need to use Recreate or make it configurable in the chart.

### DIFF
--- a/helm/templates/deployment-vllm-multi.yaml
+++ b/helm/templates/deployment-vllm-multi.yaml
@@ -120,6 +120,9 @@ spec:
         {{-   toYaml . | nindent 8 }}
       {{-   end }}
       {{- end }}
+      
+      strategy:
+        type: Recreate
 
       {{- if .Values.servingEngineSpec.runtimeClassName }}
       runtimeClassName: {{ .Values.servingEngineSpec.runtimeClassName }}


helm/templates/deployment-vllm-multi.yaml
BrianPark314, line 1: This change may be redundant because the helper already injects a strategy block (hard-coded to a RollingUpdate with specific surge and unavailable values). 
In `_helpers.tpl`
```
{{/*
Define deployment strategy
*/}}
{{- define "chart.strategy" -}}
strategy:
  rollingUpdate:
    maxSurge: 100%
    maxUnavailable: 0
{{- end }}
```
A more suitable approach would be modifying this to be value-driven so that users may control the update strategy via the valu files.

BrianPark314:
> RollingUpdates can't be used if replica count is set to 1. It would wait indefinitely to re-deploy the pods. Need to use Recreate or make it configurable in the chart.

Rolling update should work even if the replica count is set to 1. The indefinite wait is likely due to lack of additional gpu nodes.

amasolov:
@BrianPark314 there are 2 more available nodes but I see where you are coming from. I'll close this one and investigate further.

PR #127, opened 2025-02-14
Fix the problem of 1k prompts.

ggaaooppeenngg:
httpx client has a default max_connections(100) which will block the requests if the concurrency is over 100.

This commit enhances the router's concurrency capabilities. Key changes:
1. Set `httpx.AsyncClient` max connections to unlimited to unblock the requests.
2. In `process_request`, overwrite host header for proxies routing by host.
3. Call `set_ulimit` in `main` to prevent uvicorn from dropping requests due to high FD count.

### DIFF
--- a/src/vllm_router/router.py
+++ b/src/vllm_router/router.py
@@ -4,6 +4,7 @@ import threading
 import time
 import uuid
 from contextlib import asynccontextmanager
+from urllib.parse import urlparse
 
 import uvicorn
 from fastapi import FastAPI, Request, UploadFile
@@ -23,7 +24,7 @@ from vllm_router.service_discovery import (
     InitializeServiceDiscovery,
     ServiceDiscoveryType,
 )
-from vllm_router.utils import validate_url
+from vllm_router.utils import set_ulimit, validate_url
 
 httpx_client_wrapper = HTTPXClientWrapper()
 logger = logging.getLogger("uvicorn")
@@ -56,10 +57,13 @@ async def process_request(
     GetRequestStatsMonitor().on_new_request(backend_url, request_id, start_time)
 
     client = httpx_client_wrapper()
+    header_dict = dict(header)
+    # Overwrite the host if the backend server is behind a reverse proxy that routes requests based on the host.
+    header_dict["host"] = urlparse(backend_url).netloc
     async with client.stream(
         method=method,
         url=backend_url + endpoint,
-        headers=dict(header),
+        headers=header_dict,
         content=body,
         timeout=None,
     ) as backend_response:
@@ -133,6 +137,7 @@ async def route_general_request(request: Request, endpoint: str):
         stream_generator,
         status_code=status_code,
         headers={key: value for key, value in headers.items()},
+        media_type="text/event-stream",
     )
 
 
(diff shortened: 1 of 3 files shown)


gaocegege:
Thanks for your contribution! :tada: :+1:

Please fix the pre commit linting issues.

ggaaooppeenngg:
@gaocegege Done!

gaocegege (dismissed):

src/vllm_router/router.py
```diff
@@ -133,6 +137,7 @@ async def route_general_request(request: Request, endpoint: str):
         stream_generator,
         status_code=status_code,
         headers={key: value for key, value in headers.items()},
+        media_type="text/event-stream",
```
ApostaC, line 137: Why setting to "event-stream" here? If the request is not a streaming request, will the media-type still be "event-stream"?

src/vllm_router/router.py
```diff
@@ -56,10 +57,13 @@ async def process_request(
     GetRequestStatsMonitor().on_new_request(backend_url, request_id, start_time)
 
     client = httpx_client_wrapper()
+    header_dict = dict(header)
+    # Overwrite the host if the backend server is behind a reverse proxy that routes requests based on the host.
+    header_dict["host"] = urlparse(backend_url).netloc
```
ApostaC: Can you elaborate the details about the reverse proxy case?

ApostaC (commented):
Thanks for the contributions! Otherwise LGTM!

src/vllm_router/router.py
ggaaooppeenngg: This setting is for the Content-Type for the response, as the StreamingResponse in the API is always `text/event-stream`.

src/vllm_router/router.py
ggaaooppeenngg: Suppose we have two vLLM instances named InstanceA and InstanceB. To manage incoming requests to these instances, we use a general proxy, such as Nginx or another loadbalancer. This proxy is configured to route requests based on the `Host` header in the HTTP requests. For instance, requests with the `Host` header set to `region-1.example.com` are directed to InstanceA, and those with the `Host` header set to `region-2.example.com` are sent to InstanceB.

Let's assume that the router has a listening address of `127.0.0.1:30000`. However, if clients send requests directly to this address without modifying the `Host` header, the proxy will be unable to match the incoming requests to the predefined host names (`region-1.example.com` and `region-2.example.com`). As a result, the routing mechanism based on the `Host` header will not function as expected, leading to 404.

src/vllm_router/router.py
ApostaC: Ah, I see. In that case, maybe we want to change the return type to JSONResponse for non-streaming requests in the future. Let's just ignore it for now

src/vllm_router/router.py
ApostaC: Just wanted to see if I understand it correctly --- the setup is "router --> general proxy (forwarding the requests based on the "host" field) --> vllm instances", right?

In that case, I would gently push back this modification. The "general proxy" is not in the production stack's default setting, so I feel that adding the code to the default path for a non-default component may create some confusion.

As an alternative, let's think about a more extendible interface for the router to achieve this functionality? I'm open to proposals.

ggaaooppeenngg: I reverted the change, maybe leave it an issue or proposal later.

ApostaC (approved):
LGTM!

PR #155, opened 2025-02-19
[router] setuptools_scm to support version argument

### LINKED ISSUE #87: chore: Use `setuptools_scm` to replace `STACK_VERSION` in the router
Currently, we have a hard-coded version in the router.

https://github.com/vllm-project/production-stack/blob/c8cdf9de34b21532089146148efafefbec84d46e/src/vllm_router/router.py#L31

We can utilize setuptools_scm to eliminate the hard-coded string and dynamically output the pip package version.

For reference, you can check the vLLM implementation [here](https://github.com/vllm-project/vllm/blob/main/setup.py#L482).

gaocegege:
FIX #87 

```
vllm-router --version
vllm-router 0.0.9.dev2+g89305e3.d20250219
```


ApostaC (approved):
LGTM! This is very useful!

PR #298, opened 2025-03-17
Hotfix readthedocs build

hmellor:
The docs build had some issues which made it incompatible with readthedocs.

This PR resolves them.


hmellor:
Failure is unrelated (the self-hosted runner is failing to clone the repo).

Additionally, the helm workflow only ran becaue `conf.py` is a Python file.

PR #314, opened 2025-03-21
[Feat] Add nodeSelectorTerms for router & cacher servers

kinoute:
Add nodeSelectorTerms to router and cache servers to select on which nodes these deployments should be put on. Right now only the "model"/GPU deployment has this possibility.

---

- [X] Make sure the code changes pass the [pre-commit](../CONTRIBUTING.md) checks.
- [X] Sign-off your commit by using <code>-s</code> when doing <code>git commit</code>
- [X] Try to classify PRs for easy understanding of the type of changes, such as `[Bugfix]`, `[Feat]`, and `[CI]`.

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -302,6 +302,9 @@ routerSpec:
     #    hosts:
     #      - vllm-router.local
 
+  # The node selector terms to match the nodes
+  nodeSelectorTerms: []
+
   # -- TODO: Readiness probe configuration
   #startupProbe:
   #  # -- Number of seconds after the container has started before startup probe is initiated
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -303,6 +303,13 @@ routerSpec:
     #      - vllm-router.local
 
   # The node selector terms to match the nodes
+  # Example:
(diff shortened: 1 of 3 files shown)


YuhanLiu11:
@kinoute Thanks for the PR! I'll take a look soon.

helm/values.yaml
```diff
@@ -302,6 +302,9 @@ routerSpec:
     #    hosts:
     #      - vllm-router.local
 
+  # The node selector terms to match the nodes
```
YuhanLiu11, line 305: Can you give an example of the node selector term? 
Below is an example from previous fields. 
image

kinoute: @YuhanLiu11 Sure, I just updated the PR with an example

YuhanLiu11 (approved):
LGTM! Thanks for your contribution!

zevisert:
Thanks for this @kinoute - We're looking forward to the next release, especially since only x86_64 / amd64 images are built - we have no way to prevent the router from running on nodes with the wrong CPU architecture without this!

PR #319, opened 2025-03-24
feat: propose gpu and cache aware controller design

rootfs:
Since there are many ways to setup vllm and router with different routing, cache, and NLP filters, I believe it is time to expose such practices as Kubernetes CRDs so that we can use controllers to orchestrate the entire e2e flow. Although the existing helm charts can fulfill certain purposes (i.e. bootstrapping vllm and router), they lack the intelligence in scheduling, scaling, and failover. 

I understand that some of the concepts are covered in projects such as Kubernetes Gateway API (inference extension), and other projects, it would be great if we can sync on these and have a reusable flow.

### DIFF
--- a/src/router-controller/docs/gpu-aware-controller-design.md
+++ b/src/router-controller/docs/gpu-aware-controller-design.md
@@ -1,0 +1,187 @@
+# GPU and Cache Aware Inference Controller Design
+
+## Overview
+
+The Inference Controller is a Kubernetes operator that manages the routing and orchestration of vLLM inference services. It provides a declarative way to manage model serving, caching, and routing in a Kubernetes environment.
+
+## Use Cases
+
+1. **Model Serving Orchestration**
+   - Deploy and manage vLLM inference services
+   - Configure tensor parallelism for distributed model serving
+
+2. **Caching Layer Management**
+   - Deploy and manage KV cache, attention cache, and hybrid cache services
+   - Configure cache size, TTL, and transfer policies
+   - Monitor cache performance and statistics
+
+3. **Intelligent Request Routing**
+   - Route requests to appropriate model instances using routing rules
+   - Implement session affinity for consistent model access
+
+4. **Performance Optimization**
+   - Enable speculative decoding for faster inference
+   - Support prefill-decoding disaggregation
+   - Implement efficient cache transfer policies
+
+## Custom Resource Definitions (CRDs)
+
+### InferenceGateway
+
+The `InferenceGateway` CRD manages the routing layer for model inference requests with GPU-aware and cache-efficient routing.
+
+```yaml
+apiVersion: production-stack.vllm.ai/v1alpha1
+kind: InferenceGateway
+metadata:
+  name: llm-gateway
+spec:
+  schedulingPolicy: "gpu-load-aware"       # Use GPU metrics for routing
+  routeStrategy: "PrefixHash"              # Enable prefix-based routing for cache affinity
+  sessionAffinity: true                    # Ensure multi-turn sessions are pinned
+  routes:
+    - model: "llama3-70b"
+      inferenceServiceRef: "llama3-chatbot-service"
+    - model: "mistral3"
+      inferenceServiceRef: "mistral-agent-service"
+```
+
+### InferenceService
+
+The `InferenceService` CRD manages individual model serving instances with integrated caching and NLP filtering.
+
+```yaml
+apiVersion: production-stack.vllm.ai/v1alpha1
+kind: InferenceService
+metadata:
+  name: llama3-chatbot-service
+spec:
+  modelName: "llama3-70b"
+  backendRef: "chatbot-service-backend"
+  inferenceCacheRef: "llama3-chatbot-cache"
+  nlpFilters:
+    semanticCache:
+      storeRef: "redis-nlp-cache"
+      threshold: 0.85
+      ttlSeconds: 3600
+    promptGuard:
+      blockedPatterns:
+        - "Name"
+        - "SSN"
+```
+
+### InferenceCache
+
+The `InferenceCache` CRD manages GPU KV caching with support for PDD and speculative decoding.
+
+```yaml
+apiVersion: production-stack.vllm.ai/v1alpha1
+kind: InferenceCache
+metadata:
+  name: llama3-chatbot-cache
+spec:
+  modelName: "llama3-70b"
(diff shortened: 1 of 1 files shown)


src/router-controller/docs/gpu-aware-controller-design.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Custom Resource Definitions (CRDs)
+
+### InferenceGateway
+
+The `InferenceGateway` CRD manages the routing layer for model inference requests with GPU-aware and cache-efficient routing.
+
+```yaml
+apiVersion: production-stack.vllm.ai/v1alpha1
+kind: InferenceGateway
```
kfswain, line 35: Is this going to be a GW proxy built by prod-stack?

src/router-controller/docs/gpu-aware-controller-design.md
```diff
@@ -0,0 +1,187 @@
...
+
+```yaml
+apiVersion: production-stack.vllm.ai/v1alpha1
+kind: PrefillDecodingDisaggregation
+metadata:
+  name: pdd-llama3-chatbot
+spec:
+  modelName: "llama3-70b"
+  topologyHint:
+    nodeSelector:
```
ahg-g, line 117: where will this nodeSelector be used? is the plan that the controller will actually create the deployment and inject this selector? if so, what about all the other parameters in both DeploymentSpec and PodSpec, how can the user set them?

src/router-controller/docs/gpu-aware-controller-design.md
```diff
@@ -0,0 +1,187 @@
...
+spec:
+  modelName: "llama3-70b"
+  kvCacheTransferPolicy:
+    thresholdHitRate: 0.8
+    evictionThreshold: 0.9
+  pddRef: "pdd-llama3-chatbot"
+  sdRef: "sd-llama3-chatbot"
+```
+
+### SpeculativeDecoding
```
YuhanLiu11, line 91: Why are we adding Speculative Decoding to the InferenceCache CRD? Speculative Decoding is unrelated to KV cache if I understand correctly. 
Does the cache service need to know about the speculative decoding configuration in your design?

src/router-controller/docs/gpu-aware-controller-design.md
```diff
@@ -0,0 +1,187 @@
+# GPU and Cache Aware Inference Controller Design
+
+## Overview
+
+The Inference Controller is a Kubernetes operator that manages the routing and orchestration of vLLM inference services. It provides a declarative way to manage model serving, caching, and routing in a Kubernetes environment.
```
YuhanLiu11, line 5: A higher level question: Will this new API and controller design replace the Helm chart we currently use for deployment?

YuhanLiu11, line 5: Another higher level question I have is: will this new API design mean major changes to how our current router is implemented?

src/router-controller/docs/gpu-aware-controller-design.md
```diff
@@ -0,0 +1,187 @@
...
+The `InferenceGateway` CRD manages the routing layer for model inference requests with GPU-aware and cache-efficient routing.
+
+```yaml
+apiVersion: production-stack.vllm.ai/v1alpha1
+kind: InferenceGateway
+metadata:
+  name: llm-gateway
+spec:
+  schedulingPolicy: "gpu-load-aware"       # Use GPU metrics for routing
+  routeStrategy: "PrefixHash"              # Enable prefix-based routing for cache affinity
```
YuhanLiu11, line 40: Are these "gpu-load-aware" and "PrefixHash" two different algorithms for routing? If so, should we keep only one of them in this CRD?

src/router-controller/docs/gpu-aware-controller-design.md
```diff
@@ -0,0 +1,187 @@
...
+The `InferenceCache` CRD manages GPU KV caching with support for PDD and speculative decoding.
+
+```yaml
+apiVersion: production-stack.vllm.ai/v1alpha1
+kind: InferenceCache
+metadata:
+  name: llama3-chatbot-cache
+spec:
+  modelName: "llama3-70b"
+  kvCacheTransferPolicy:
```
YuhanLiu11, line 84: What is this ``kvCacheTransferPolicy`` referring to specifically? Sounds like eviction policy, given the fields ``evictionThreshold``. If so, we should use another name for this.

src/router-controller/docs/gpu-aware-controller-design.md
rootfs: nodeSelector provides an informed estimate of how to best utilize the underlying network topology for low latency GPU2GPU transfer. 

This is a minimal set of requirement, more info are surely to be investigated.

src/router-controller/docs/gpu-aware-controller-design.md
rootfs: Right, SD is not strictly a KV cache optimization, suggestion of better terms is appreciated.

src/router-controller/docs/gpu-aware-controller-design.md
rootfs: The idea is to set a KV cache transfer policy, the fields are really based on my prior cache knowledge, if they don't make sense here, please share more thoughts.

src/router-controller/docs/gpu-aware-controller-design.md
rootfs: sounds good, will fix.

src/router-controller/docs/gpu-aware-controller-design.md
rootfs: The idea here is to use existing Production Stack for an in-cluster gateway PoC (since there is no extproc integration yet). If this goes to another environment, i.e. Kubernetes Gateway, an Envoy proxy based setting with existing GatewayClass API is potentially the way to go.

src/router-controller/docs/gpu-aware-controller-design.md
rootfs: Quite the opposite:D I am trying to reuse the existing helm chart as the blueprint for the new API design.

YuhanLiu11:
@rootfs @ahg-g @kfswain I implemented an experimental version of the inference engine controller at PR #348, feel free to comment! By the way this is only a design example (not the final design).

PR #320, opened 2025-03-25
[Bugfix] Update logger handler to handle stdout/stderr prorperly

corona10:
* [logging.StreamHandler](https://docs.python.org/3/library/logging.handlers.html#logging.StreamHandler) emits the log as stderr by default.
* It will be the issue at the production logging system where stderr as the system error.
* In our production, the logging detect system recognizes it as the error status since it emits stderr even if the log level is info.
* To fix it properly, there should be two logging.StreamHandler. One is for stdout the other is for the stderr.
* In this change.
    - logging.StreamHandler for stdout only emit until the log level is INFO.
    - If the log level is greater than WARN, then the logger will emit the log as stderr.
* And the other fix is propagate, if the propagate is set as True, in real logging system, it detects duplicated log from unicorn. So to prevent propagation to unicorn log, propagate should be set as False.

### DIFF
--- a/src/vllm_router/log.py
+++ b/src/vllm_router/log.py
@@ -1,4 +1,5 @@
 import logging
+import sys
 from logging import Logger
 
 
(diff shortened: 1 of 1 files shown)


corona10:
Gentle ping @Shaoting-Feng @YuhanLiu11

YuhanLiu11 (approved):
LGTM! Thanks for the fix.

PR #367, opened 2025-04-14
[Bugfix] added model name to deployment selector

TamKej:
Added model label to deployment selector.

In order to ensure the deployments for each model select only the pods associated with their model, I added the model label to the selector.


YuhanLiu11 (approved):
LGTM!

PR #395, opened 2025-05-01
[Router] add sentry sdk and cli args

pwuersch:
We want to deploy the router part of production stack in our own environment. We make use of [Sentry](https://sentry.io) for error reporting. This PR adds the `sentry-sdk` package along with the required extras to make use of Sentry's integration with FastAPI and HTTPX. The setup will only be done in cases where the new `--sentry-dsn` flag has been set at startup. Otherwise, the router will start without any changes.

I tried my best to assume the configuration of the `pyproject.toml` and `requirements.txt` dependencies. Please let me know if I should change anything regarding the setup. Other than that, the Sentry SDK doesn't require much setup other than the call to `sentry_sdk.setup()` which monkey-patches some functions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,6 +19,7 @@ dependencies = [
     "numpy==1.26.4",
     "prometheus-client==0.21.1",
     "python-multipart==0.0.20",
+    "sentry-sdk[fastapi,httpx]==2.27.0",
     "uhashring==2.3",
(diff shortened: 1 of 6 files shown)


YuhanLiu11:
@pwuersch thanks for your PR! I'll take a look soon.

YuhanLiu11 (approved):
LGTM!

Issue #389, opened 2025-04-29
feature: add model aliases


max-wittig:
Previously we used [litellm](https://github.com/BerriAI/litellm) and there we could define model aliases.

See the [documentation in litellm](https://docs.litellm.ai/docs/completion/model_alias)

This can be used, for example to map known commercial model names to local ones or route older models to newer more performant ones to preserve compatibility in existing configurations of end users.

E.g. this was our configuration so when a user requested `gpt-4`, the request would just be served by `mistral-7b-instruct` in the background.

```yaml
router_settings:
  model_group_alias:
    gpt-4: mistral-7b-instruct
    gpt-3.5-turbo: mistral-7b-instruct
    gpt-3.5-turbo-16k: mistral-7b-instruct
    gpt-4-32k: mistral-7b-instruct
    codellama-13b-instruct: mistral-7b-instruct
    text-embedding-3-small: bge-m3
    text-embedding-3-large: bge-m3
```

In addition it was possible to hide these aliases in the `v1/models` models endpoint.

I think that would be a great addition. Thanks for your consideration!

### Why do you need this feature?

* preserve compatibility of model names when newer models are released
* allow users to specify known commercial models and route the requests to open weight models instead

/cc @nejch

YuhanLiu11:
@max-wittig this would be a great feature to have! Would you like to take a stab at implementing this feature? Thanks!

max-wittig:
@YuhanLiu11 Sure, I'm on it!

BrianPark314:
I think vllm already supports model alias as runtime via `--served-model-name` argument.
https://docs.vllm.ai/en/v0.5.5/models/engine_args.html

nejch:
> I think vllm already supports model alias via `--served-model-name` engine runtime argument.
> https://docs.vllm.ai/en/v0.5.5/models/engine_args.html

We looked at that initially but it's a bit clunky - it's much easier to add arbitrary aliases on the proxy level rather than sort of faking it on individual workers, at least for our use case. Also decouples the deployment a bit more this way.

max-wittig:
Thanks for the merge!

PR #397, opened 2025-05-01
[Feat][Router] add model aliases

max-wittig:
This feature allows:
* to preserve compatibility of model names when newer models are released
* allow users to specify known commercial models and route the requests to open weight models instead

E.g. if I want to replace `mistral-7b-instruct` with the more powerful `mistral-nemo-instruct-2407` and not disrupt the users, and at the same time route `gpt-4` to `mistral-nemo-instruct-2407` I could just specify it like this

```sh
python3 -m vllm_router.app \
--static-aliases "mistral-7b-instruct:mistral-nemo-instruct-2407,gpt-4:mistral-nemo-instruct-2407"
```

FIX https://github.com/vllm-project/production-stack/issues/389

---

- [x] Make sure the code changes pass the [pre-commit](../CONTRIBUTING.md) checks.

<details>

<ul>
</ul>

<h3>Code Quality</h3>

<ul>
</ul>

, Shaoting-Feng or ApostaC.

### DIFF
--- a/docs/requirements-docs.txt
+++ b/docs/requirements-docs.txt
@@ -6,3 +6,4 @@ sphinx-copybutton==0.5.2
 sphinx-togglebutton==0.3.2
 sphinx_design==0.6.1
 sphinxemoji==0.3.1
+starlette==0.45.3
--- a/src/vllm_router/services/request_service/request.py
+++ b/src/vllm_router/services/request_service/request.py
@@ -27,6 +27,7 @@ from vllm_router.services.request_service.rewriter import (
     get_request_rewriter,
     is_request_rewriter_initialized,
 )
+from vllm_router.utils import replace_model_in_request_body, update_content_length
 
 try:
     # Semantic cache integration
@@ -195,11 +196,18 @@ async def route_general_request(
             logger.warning("Failed to parse rewritten request body as JSON")
 
     # TODO (ApostaC): merge two awaits into one
-    endpoints = get_service_discovery().get_endpoint_info()
+    service_discovery = get_service_discovery()
+    endpoints = service_discovery.get_endpoint_info()
     engine_stats = request.app.state.engine_stats_scraper.get_engine_stats()
     request_stats = request.app.state.request_stats_monitor.get_request_stats(
         time.time()
     )
+
+    aliases = getattr(service_discovery, "aliases", None)
+    if aliases and requested_model in aliases.keys():
+        requested_model = aliases[requested_model]
+        request_body = replace_model_in_request_body(request_json, requested_model)
+        update_content_length(request, request_body)
 
     endpoints = list(filter(lambda x: x.model_name == requested_model, endpoints))
     if not endpoints:
--- a/README.md
+++ b/README.md
@@ -115,6 +115,7 @@ The router ensures efficient request distribution among backends. It supports:
 - Routing to endpoints that run different models
 - Exporting observability metrics for each serving engine instance, including QPS, time-to-first-token (TTFT), number of pending/running/finished requests, and uptime
 - Automatic service discovery and fault tolerance by Kubernetes API
+- Model aliases
 - Multiple different routing algorithms
   - Round-robin routing
   - Session-ID based routing
--- a/src/tests/test_utils.py
+++ b/src/tests/test_utils.py
@@ -1,0 +1,63 @@
+import json
+from unittest.mock import MagicMock
+
+import pytest
+from starlette.datastructures import MutableHeaders
+
+from vllm_router import utils
+
+
+@pytest.mark.parametrize(
+    "aliases,expected_result",
+    (
+        ("gpt-4:mistral-nemo-instruct-2407", {"gpt-4": "mistral-nemo-instruct-2407"}),
+        (
+            "gpt-4:mistral-nemo-instruct-2407,gpt-3.5:mistral-nemo-instruct-2407",
+            {
+                "gpt-4": "mistral-nemo-instruct-2407",
+                "gpt-3.5": "mistral-nemo-instruct-2407",
+            },
+        ),
+        (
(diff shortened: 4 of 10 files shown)


src/vllm_router/services/request_service/request.py
```diff
@@ -194,12 +195,19 @@ async def route_general_request(
...
     engine_stats = request.app.state.engine_stats_scraper.get_engine_stats()
     request_stats = request.app.state.request_stats_monitor.get_request_stats(
         time.time()
     )
 
+    aliases = getattr(service_discovery, "aliases", None)
+    if aliases and requested_model in aliases.keys():
+        requested_model = aliases[requested_model]
+        request_body = replace_model_in_request_body(request, request_body, request_json, requested_model)
+        update_content_length(request, request_body)
```
max-wittig, line 210: I doubt that the `request_rewriter` above works without this part.

max-wittig:
Before I add documentation, tests etc. Could you just give it a brief review just that it goes in the right direction?

@YuhanLiu11

YuhanLiu11:
@max-wittig Thanks for the PR! A higher level question, could you elaborate on why using vLLM’s ``--served-model-name`` doesn’t work for your use case?

max-wittig:
@YuhanLiu11 Our deployment runs on multiple machines with different kind of configurations. The GPU instances are real physical machines in a data center, some on AWS etc, which are harder to adjust and update than just adjusting the vllm-router. It allows us to have the configuration deployed in one place and be effective.

Basically its nice for faster iterations and it seems more natural to have this part inside the router.

YuhanLiu11:
@max-wittig I see. In that case this PR looks good on a high level (if it can pass the tests).

max-wittig:
@YuhanLiu11 Thanks for the comments! I will fix the tests and add some for the functionality then!

max-wittig:
I didn't add tests for the service router as currently there are none yet, but I'm open to some in general in a follow-up.

YuhanLiu11:
@max-wittig thanks! I'll take a closer look at this PR later today or tomorrow (since this is a relatively large one)!

docs/requirements-docs.txt
```diff
@@ -6,3 +6,4 @@ sphinx-copybutton==0.5.2
 sphinx-togglebutton==0.3.2
 sphinx_design==0.6.1
 sphinxemoji==0.3.1
+starlette==0.45.3
```
pwuersch, line 9: @max-wittig is adding this package here required? Firstly, it's a change in the documentation requirements, and secondly, `starlette` is already included when installing `fastapi`, see https://github.com/fastapi/fastapi/blob/9a33ba46ac3a3ef0b845c8515822f4e13c0fb443/pyproject.toml#L46. I believe `fastapi` is already installed in all actions where the tests are run. Or was this due to some other change / failing CI step?

max-wittig: yes. Otherwise the pipeline breaks.

src/vllm_router/services/request_service/request.py
```diff
@@ -195,12 +196,19 @@ async def route_general_request(
...
     engine_stats = request.app.state.engine_stats_scraper.get_engine_stats()
     request_stats = request.app.state.request_stats_monitor.get_request_stats(
         time.time()
     )
 
+    aliases = getattr(service_discovery, "aliases", None)
+    if aliases and requested_model in aliases.keys():
+        requested_model = aliases[requested_model]
+        request_body = replace_model_in_request_body(request_json, requested_model)
+        update_content_length(request, request_body)
```
pwuersch, line 210: This double-lookup into `aliases` could be avoided. WDYT? It is a hot-path that is executed on every request, so this might be worth optimizing although the lookup is not that expensive.

```suggestion
    if aliases:
        if actual_model := aliases.get(requested_model):
            request_body = replace_model_in_request_body(request_json, actual_model)
            update_content_length(request, request_body)
```

pwuersch (changes requested):
@max-wittig I went over your changes and found one small improvement that could be done. Other than that, the feature looks good to me and solves our use-case as expected. Thank you very much for implementing it.

src/vllm_router/services/request_service/request.py
max-wittig: in my opinion this makes it a bit harder to read, but I will leave it up to the maintainers.

pwuersch: agree 👍🏻

src/vllm_router/services/request_service/request.py
```diff
@@ -195,12 +196,19 @@ async def route_general_request(
             logger.warning("Failed to parse rewritten request body as JSON")
 
     # TODO (ApostaC): merge two awaits into one
-    endpoints = get_service_discovery().get_endpoint_info()
+    service_discovery = get_service_discovery()
+    endpoints = service_discovery.get_endpoint_info()
```
YuhanLiu11, line 200: Why do we need to split this line into two?

src/vllm_router/services/request_service/request.py
YuhanLiu11: The current code looks good to me

src/vllm_router/services/request_service/request.py
max-wittig: Because I need to use the service discovery variable later in line 206 to get the aliases, if they exist:

https://github.com/vllm-project/production-stack/pull/397/files#diff-3e852cfcf107dfd78ee2b7c7e88e05428b7f2bffcfc5ca1dce0c070caf563f2aR206

Let me know, if there is a better way to do things.

YuhanLiu11: I see. ok. This looks good then.

YuhanLiu11 (approved):
LGTM, thanks for the feature!

PR #423, opened 2025-05-12
[Misc] use schema https://json-schema.org/draft/2020-12/schema

sh1ng:
The current version uses a dynamic version that redirects to the latest version and may change over time. It also breaks CUE logic when dealing with the existing configuration.

Two issues provide more details and explain the necessity of the changes.

- https://github.com/cue-lang/cue/issues/3461
- https://github.com/argoproj/argo-workflows/pull/14092

### DIFF
--- a/helm/values.schema.json
+++ b/helm/values.schema.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "http://json-schema.org/schema#",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
   "type": "object",
   "properties": {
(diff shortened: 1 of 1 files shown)


YuhanLiu11:
@sh1ng thanks for the quick fix! Can you fix the DCO issue?

sh1ng:
@YuhanLiu11 done!

Also, I found a few inconsistencies between the provided JSON schema and a few examples. But they will be in a separate PR.

YuhanLiu11 (approved):
LGTM! Thanks

PR #427, opened 2025-05-13
[Misc] fix values.schema.json so it does not contradict to examples from tutorials/assets/*.yaml

sh1ng:
While working on the unification of the configuration process by using [cue](https://cuelang.org/), I had a few errors. 
This PR fixes them.

You can convert `values.schema.json` to cue format first by running
```
cue import helm/values.schema.json
```
and then verify existing configs
```
cue vet -i helm/values.schema.cue tutorials/assets/*.yaml
```

It also includes https://github.com/vllm-project/production-stack/pull/423.


sh1ng:
@YuhanLiu11 could you review it please?

PR #464, opened 2025-05-28
[Feat]: Add extraVolumeMount for initContainer

JasmondL:
This PR adds extraVolumeMount to the initContainer to facilitate the download of models or metadata to specific volume mounts. 

This PR also includes changes to allow a list of imagePullSecrets. The reason behind this is that the initContainer and the main container may come from different repositories.

### DIFF
--- a/helm/templates/_helpers.tpl
+++ b/helm/templates/_helpers.tpl
@@ -146,6 +146,15 @@ limits:
 {{- end }}
 
 {{/*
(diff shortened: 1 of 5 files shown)


YuhanLiu11:
Hey @JasmondL Thanks for the PR! Could you elaborate why code changes related to pod annotations are also included in this PR?

PR #566, opened 2025-07-03
[Router] Request Load Balancing, KVAware Router

allytotheson:
## Summary

This PR introduces `LoadBalancingRouter`, a routing strategy that prioritizes Quality of Experience (QoE) by:

1. **Preferring endpoints with KV cache hits** to reuse computation and reduce latency.
2. Falling back to **TTFT estimation + load balancing** when no cache hit is available.

TTFT is estimated using:
- Prompt length
- Model parameter size (inferred from endpoint name, e.g., "7B", "13B")
- Current endpoint load
- Hardware throughput (FLOPS) and memory bandwidth (HBM)

## Key Changes

- Added `LoadBalancingRouter` class with:
  - KV cache-aware routing logic via `LMCacheControllerManager`
  - TTFT estimation based on model and prompt characteristics
  - Load tracking and balancing across endpoints
- Automatically infers model size from endpoint names

## Notes

- KV cache hits take precedence over TTFT/load heuristics.
- Call `complete_request(endpoint_url)` after each request finishes to properly track load.
- Uses default `model_params` if inference from name fails.

### DIFF
--- a/src/tests/requirements.txt
+++ b/src/tests/requirements.txt
@@ -3,3 +3,4 @@ httpx
 openai
 uvicorn
 vllm
+lmcache


Shaoting-Feng:
What is the difference between this routing logic and `KVAWARE = "kvaware"`?

allytotheson:
> What is the difference between this routing logic and `KVAWARE = "kvaware"`?

This router falls back to estimating TTFT and balancing request loads across endpoints when KV cache matches are insufficient. The TTFT algorithm allows for the best matching endpoint to be selected.

PR #626, opened 2025-08-02
[Feat] Router-side queuing support

allytotheson:
## Summary

This PR introduces an asynchronous `EndpointQueueManager` to handle request queueing and dispatch for LLM inference endpoints. It enables queue-based request routing to improve Quality of Experience (QoE), particularly under high load, while supporting immediate dispatch when endpoints are free.

## Key Features

- **Per-endpoint request queues** with `PriorityQueue` and `asyncio.Condition`
- **Scheduler loop per endpoint** that handles dispatch based on availability
- **Optional queue toggling** via a configuration flag
- **Rerouting of stale requests** that exceed a configurable wait threshold
- **Scaffolding** in place for future session aware and/or KV cache integration when handling stale requests
- **Graceful shutdown** of endpoint schedulers

## Highlights

- Requests are either:
  - Enqueued if the endpoint is overloaded
  - Immediately dispatched if the endpoint is free or queueing is disabled
- Rerouting logic kicks in for requests that exceed `max_queue_wait_time`

## Future Work

- Implement KV cache-aware, session-based stale-request rerouting
- Knapsack-style group dispatch logic for optimized batch handling


ruizhang0101:
Hi, I am wondering the difference between it and vllm scheduler. There is already a request scheduler inside vllm, why do we need this?

allytotheson:
> Hi, I am wondering the difference between it and vllm scheduler. There is already a request scheduler inside vllm, why do we need this?

Hi, thanks for the question! Is the vllm scheduler you're referring to the one that manages token-level execution at the backend? 

This queueing system focuses on admission control at the router level, before the requests are sent to the backend. The main goal is to avoid overwhelming any single backend. I have default thresholds like 10 concurrent requests and 95% GPU usage as signals to hold off on dispatching new requests.
What makes this useful is that it enables strategies like rerouting requests to another endpoint if they’ve been stuck in a queue too long. It also opens the door for more advanced dispatch logic — for example, priority-based queueing (even though currently everything is treated as high priority).

That said, I'm definitely open to exploring smarter strategies that don’t strictly block requests when thresholds are hit to allow it to work better with the request scheduler.

src/vllm_router/parsers/parser.py
```diff
@@ -339,6 +339,32 @@ def parse_args():
         help="The threshold for kv-aware routing.",
     )
 
+    parser.add_argument(
+        "--enable_queue",
+        action=argparse.BooleanOptionalAction,
+        default=True,
```
ruizhang0101: Can you set default to false since this feature is not totally finished?

ruizhang0101 (changes requested):
I think this one is a super good feature! Thanks for contributing! Could you make the enable_queue default value false? And then I think we are good to go for now. Hoping to see the following changes.

lucas-tucker:
Has this been tested and has shown to improve performance? Router-side queuing makes sense, but this appears to be more of a priority mechanism on the _endpoints_ as `route_general_request` is not touched.

ruizhang0101 (approved):
LGTM

ikaadil:
@allytotheson @zerofishnoodles do you have any update for this PR?

ruizhang0101:
> @allytotheson @zerofishnoodles do you have any update for this PR?

There are some conflicts right now that needs to be resolved.

punitvara:
Hi @allytotheson — really like the EndpointQueueManager design here, and it's great that it already has an LGTM. Since it's been stalled on merge conflicts for a while and router-side queuing is still on the 2026 roadmap (https://github.com/vllm-project/production-stack/issues/855) (P1), I wanted to check: are you still planning to pick this back up?

If not, I'd be happy to rebase and carry it across the line — I recently got the priority routing PR (#996) merged into the router, so I'm already up to speed on this part of the codebase and would try to stay close to your original design/approach rather than re-architecting.

No pressure either way, just don't want to duplicate effort if you're already on it. 
cc: @ruizhang0101

Issue #300, opened 2025-03-17
[Roadmap] vLLM Production Stack roadmap for 2025 Q2


YuhanLiu11:
This issue document the roadmap for 2025 Q5. We will keep updating this document to include the related issues, pull requests, and discussions in the [#production-stack](https://vllm-dev.slack.com/archives/C089SMEAKRA) channel in the vLLM slack. Feel free to leave comments / questions in this issue! 

## Multi-node support
- [ ] (P0) Single-vLLM instance multi-node setup (e.g., PP on 2 different nodes)
- [ ] (P1) The reference implementation for disaggregated prefill

## K8s-native control plane

- [ ] (P0) Model CRD for vLLM deployment operations
- [ ] (P1) Router CRD for router deployment operations (#212 )
- [ ] (P1) LoRA CRD & controller (Proposal: [lora-k8s-support.md](production-stack/proposals/lora-k8s-support.md) )
- [ ] (P1) Autoscale CRD & controller (#238 )

### Router “frontend”

- [ ] (P1) Envoy ext_proc integration (Draft PR: #240) 
- [ ] (P1) Gateway API extension integration 
- [ ] (P2) Router performance enhancements
  - [ ] Nuitka compilation for the current router
  - [ ] Rust/Go/Nginx-based frontend for router 
- [ ] (P2) Integration with Application interfaces like MCP or other agentic workflows (#244)

### Router Core Logic
- [ ] (P0) Queuing support in the router
- [ ] (P2) Priority routing 
- [ ] (P2) Routing to external providers like OpenAI or Anthropic
- [ ] (P2) Request migration when the vLLM instance fails
- [ ] (P2) Router extension modules
  - [ ] Semantic caching (#210 )
  - [ ] PII detection (#235 )

## CI/CD and testing

- [ ] (P1) Comprehensive unit tests for router
- [ ] (P1) Use CPU-based vLLM for functionality tests
- [ ] (P1) Github actions for router performance benchmarking
- [ ] (P2) Github actions for building router docker images for different architectures
- [ ] (P2) Release bot to automatically release new versions (helm chart + k8s controller packages + docker images)

## Misc
- [ ] (P1) Documentation 
- [ ] (P2) Tutorials for more cloud platforms, different models, and features
- [ ] (P2) End-to-end performance benchmarking: (more workload, more setups)

ahg-g:
[...]

PR #639, opened 2025-08-11
[Router][CI/CD and misc.] Add RoundRobinRouter logic testing

lucas-tucker:
Addresses roadmap #300 

<p> Adds a basic test for RoundRobinRouter to ensure the logic is correct. In particular, performs requests to the router over different service discoveries (i.e. different number of endpoints) and ensures even distribution to the endpoint URLs along the way. </p>

<p> To run, use
<code> pytest test_roundrobin_router.py </code>
</p>


zerofishnoodles (approved):
LGTM

PR #643, opened 2025-08-12
[Feat][Router] Add simple request migration for vLLM failure

lucas-tucker:
Addresses the Q2 roadmap #300

Adds a simple loop in the `route_general_request` function in `request.py` to ensure that a number of retries take place if streaming from the selected url results in failure.

### DIFF
--- a/src/vllm_router/services/request_service/request.py
+++ b/src/vllm_router/services/request_service/request.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 # --- Request Processing & Routing ---
+import asyncio
 import json
 import os
 import time
@@ -136,8 +137,54 @@ async def process_request(
         )
 
 
+def perform_service_discovery(request, request_json, request_endpoint):
+    # TODO (ApostaC): merge two awaits into one
+    service_discovery = get_service_discovery()
+    endpoints = service_discovery.get_endpoint_info()
+
+    aliases = getattr(service_discovery, "aliases", None)
+    if aliases and requested_model in aliases.keys():
+        requested_model = aliases[requested_model]
+        request_body = replace_model_in_request_body(request_json, requested_model)
+        update_content_length(request, request_body)
+
+    if not request_endpoint:
+        endpoints = list(
+            filter(
+                lambda x: requested_model in x.model_names and not x.sleep,
+                endpoints,
+            )
+        )
+        engine_stats = request.app.state.engine_stats_scraper.get_engine_stats()
+        request_stats = request.app.state.request_stats_monitor.get_request_stats(
+            time.time()
+        )
+    else:
+        endpoints = list(
+            filter(
+                lambda x: requested_model in x.model_names
+                and x.Id == request_endpoint
+                and not x.sleep,
+                endpoints,
+            )
+        )
+        engine_stats, request_stats = None, None
+
+    if not endpoints:
+        return JSONResponse(
+            status_code=400,
+            content={
+                "error": f"Model {requested_model} not found or vLLM engine is sleeping."
+            },
+        )
+    return endpoints, engine_stats, request_stats
+
+
 async def route_general_request(
-    request: Request, endpoint: str, background_tasks: BackgroundTasks
+    request: Request,
+    endpoint: str,
+    background_tasks: BackgroundTasks,
+    attempted_reroutes: int = 0,
 ):
     """
     Route the incoming request to the backend server and stream the response back to the client.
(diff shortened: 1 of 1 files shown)


zerofishnoodles (changes requested):
Hi Lucas, I am wondering what if the vllm instance is up and can be discovered but due some other reasons it failed with the request, won't it continuously route to the same endpoint? In this scenario, the request will never succeed. will it be better to send to different ones?

lucas-tucker:
Yes. Storing endpoints as mapping from URL to object would be useful in removal. Can also go through the list if you'd like.

ruizhang0101:
> Yes. Storing endpoints as mapping from URL to object would be useful in removal. Can also go through the list if you'd like.

Can you route a different one if the current one fails? There are some issues considering about the dead pod IP #656 , in which case the request will fail continuously if re-routing.

jonoillar:
Could we have the number of attempted_reroutes configurable ? Thanks :)

lucas-tucker:
It is currently configurable as an argument passed to `route_general_request`, with default value `0`

ikaadil:
> It is currently configurable as an argument passed to `route_general_request`, with default value `0`

Could you please add in the `src/vllm_router/parsers/parser.py`? Thanks

src/vllm_router/parsers/parser.py
```diff
@@ -367,6 +367,13 @@ def parse_args():
         help="The threshold for kv-aware routing.",
     )
 
+    parser.add_argument(
+        "--request-reroutes",
```
ikaadil: You may consider renaming this argument to a more explicit name, such as `--max-instance-failover-reroute-attempts`

src/vllm_router/services/request_service/request.py
```diff
@@ -136,8 +137,58 @@ async def process_request(
         )
 
 
+def perform_service_discovery(
+    request, request_json, request_endpoint, requested_model, error_urls
+):
+    # TODO (ApostaC): merge two awaits into one
```
ikaadil: Please remove this comment. This is irrelevant now.

src/vllm_router/parsers/parser.py
lucas-tucker: Agreed

ikaadil:
Could you clarify the best way to review this PR?
Here’s what I did when testing it:

1. Installed the Helm chart:
```
helm install vllm ./helm -f .github/values-01-2pods-minimal-example.yaml
```
2. Ran the stress test script:
```
tests/e2e/stress-test.sh
```

3. After the script successfully processed about 1,000 requests, I manually deleted one of the pods.
4. As soon as that happened, the stress test failed.
The test output:
```
tests/e2e/stress-test.sh
[INFO] Checking prerequisites...
[INFO] Router stress test configuration:
[INFO]   Concurrent requests: 2000
[INFO]   Total requests: 10000
[INFO]   Router port: 30080
[INFO]   Backend ports: 8001, 8002
[INFO]   Model: facebook/opt-125m
[INFO] Starting router with round-robin routing (stress test mode)
[INFO] Router started with PID: 1100606
[INFO] Waiting for router to be ready...
[INFO] Router is ready
[INFO] Running stress test with Apache Bench
[INFO] Concurrent: 2000, Total: 10000
This is ApacheBench, Version 2.3 <$Revision: 1879490 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking localhost (be patient)
Completed 1000 requests
Completed 2000 requests
Completed 3000 requests
Completed 4000 requests
apr_pollset_poll: The timeout specified has expired (70007)
Total of 4529 requests completed
```
So it looks like deleting a pod during the test causes the router to fail.
What’s the recommended way to validate whether this behavior is expected, or if it indicates an issue with the PR?

ikaadil:
@lucas-tucker @zerofishnoodles do you have any update about this PR?

ruizhang0101:
> @lucas-tucker @zerofishnoodles do you have any update about this PR?

There are some conflicts right now that need to be resolved.

ikaadil:
> > @lucas-tucker @zerofishnoodles do you have any update about this PR?
> 
> There are some conflicts right now that need to be resolved.

@lucas-tucker hasn’t been active on this repository for some time and has recently joined Y Combinator. I’m happy to create a new PR from scratch to take care of this.

ruizhang0101:
@ikaadil Yes, that would be really helpful!  Thanks!

PR #666, opened 2025-08-26
[Bug][Router] ZombieIP: router routes to model pod that doesn't exist anymore

jonoillar:
FILL IN THE PR DESCRIPTION HERE

FIX https://github.com/vllm-project/production-stack/issues/656

**BEFORE SUBMITTING, PLEASE READ THE CHECKLIST BELOW AND FILL IN THE DESCRIPTION ABOVE**

---

- [ ] Make sure the code changes pass the [pre-commit](https://github.com/vllm-project/production-stack/blob/main/CONTRIBUTING.md) checks.
- [ ] Sign-off your commit by using <code>-s</code> when doing <code>git commit</code>
- [ ] Try to classify PRs for easy understanding of the type of changes, such as `[Bugfix]`, `[Feat]`, and `[CI]`.

<details>

<summary><b> Detailed Checklist (Click to Expand) </b></summary>

<p>Thank you for your contribution to production-stack! Before submitting the pull request, please ensure the PR meets the following criteria. This helps us maintain the code quality and improve the efficiency of the review process.</p>

<h3>PR Title and Classification</h3>
<p>Please try to classify PRs for easy understanding of the type of changes. The PR title is prefixed appropriately to indicate the type of change. Please use one of the following:</p>
<ul>
    <li><code>[Bugfix]</code> for bug fixes.</li>
    <li><code>[CI/Build]</code> for build or continuous integration improvements.</li>
    <li><code>[Doc]</code> for documentation fixes and improvements.</li>
    <li><code>[Feat]</code> for new features in the cluster (e.g., autoscaling, disaggregated prefill, etc.).</li>
    <li><code>[Router]</code> for changes to the <code>vllm_router</code> (e.g., routing algorithm, router observability, etc.).</li>
    <li><code>[Misc]</code> for PRs that do not fit the above categories. Please use this sparingly.</li>
</ul>
<p><strong>Note:</strong> If the PR spans more than one category, please include all relevant prefixes.</p>

<h3>Code Quality</h3>

<p>The PR need to meet the following code quality standards:</p>

<ul>
    <li>Pass all linter checks. Please use <code>pre-commit</code> to format your code. See <code>README.md</code> for installation.</li>
    <li>The code need to be well-documented to ensure future contributors can easily understand the code.</li>
    <li> Please include sufficient tests to ensure the change is stay correct and robust. This includes both unit tests and integration tests.</li>
</ul>

<h3>DCO and Signed-off-by</h3>
[...]

PR #668, opened 2025-08-27
[Bugfix] Fix routing to delete endpoint

### LINKED ISSUE #656: [Bug][Router] ZombieIP: router routes to model pod that doesn't exist anymore
### Describe the bug

For the service discovery `K8sPodIPServiceDiscovery`, the `self.available_engine` variable does not get updated sometimes: it still holds entries corresponding to pod that do not exist anymore. I called them ZombieIP

The consequence: some request are routed to these deleted pods. In production, we've had router pod that for *days* where routing request to model pod that do not exist anymore

### To Reproduce

# How to reproduce the issue

For this, I needed to deploy custom code.
Modification to bring to the vllm router code:

```
--- a/src/vllm_router/service_discovery.py
+++ b/src/vllm_router/service_discovery.py
@@ -386,6 +386,7 @@ class K8sPodIPServiceDiscovery(ServiceDiscovery):
         self.watcher_thread.start()
         self.prefill_model_labels = prefill_model_labels
         self.decode_model_labels = decode_model_labels
+        self.failing_counter = 0

     @staticmethod
     def _check_pod_ready(container_statuses):
@@ -502,7 +503,12 @@ class K8sPodIPServiceDiscovery(ServiceDiscovery):
         url = f"http://{pod_ip}:{self.port}/v1/models"
         try:
             headers = None
+            self.failing_counter +=1
+            logger.info(f"{self.failing_counter=}")
             if VLLM_API_KEY := os.getenv("VLLM_API_KEY"):
+                if self.failing_counter > 3:
+                    time.sleep(60)
+                    VLLM_API_KEY = "wrong_key_jklkjlkj"
                 logger.info("Using vllm server authentication")
                 headers = {"Authorization": f"Bearer {VLLM_API_KEY}"}
             response = requests.get(url, headers=headers)
@@ -570,17 +576,23 @@ class K8sPodIPServiceDiscovery(ServiceDiscovery):

         while self.running:
             try:
+                logger.info(f"K8s watcher started{self.get_endpoint_info()}")
+                logger.info("time out is 30")
+                logger.info("Jon latest version v2")
                 for event in self.k8s_watcher.stream(
                     self.k8s_api.list_namespaced_pod,
                     namespace=self.namespace,
                     label_selector=self.label_selector,
                     timeout_seconds=30,
                 ):
+
                     pod = event["object"]
                     event_type = event["type"]
                     pod_name = pod.metadata.name
                     pod_ip = pod.status.pod_ip

+                    logger.info(f"pod_name: {pod_name} pod_ip: {pod_ip} event_type: {event_type}")
+
                     # Check if pod is terminating
                     is_pod_terminating = self._is_pod_terminating(pod)
                     is_container_ready = self._check_pod_ready(
```

Modification explanation: 
After some successful calls to `/v1/models` I introduce a "blocking" call that takes more than `30` seconds (I set to 60) and finally fails (due to wrong vllm api key)

The yaml file to use for values:

```
servingEngineSpec:
  vllmApiKey: "fake_key"
  runtimeClassName: ""
  modelSpec:
  - name: "opt125m"
    repository: "vllm/vllm-openai"
    tag: "latest"
    modelURL: "facebook/opt-125m"

    replicaCount: 2

    vllmConfig:
      extraArgs:
        - "--chat-template=./examples/template_chatml.jinja"
        - "--disable-log-requests"

routerSpec:
  repository: "git-act-router"  # Use the image we just built
  imagePullPolicy: "IfNotPresent"
  strategy:
    type: Recreate
  enableRouter: true
  serviceDiscovery: "k8s"  # This enables K8sPodIPServiceDiscovery
  routingLogic: "roundrobin"
  extraArgs:
    - "--log-level"
    - "debug"
```

Note the `vllmApiKey: "fake_key"`

## Step by step issue reproduction
1. Deploy the helm chart
2. Monitor the logs of the router

You will see at first, everything looks normal. After 3 iterations of pod discovery, you will start seeing this error log:

```
[2025-08-20 10:30:44,354] ERROR: Failed to get model names from http://10.244.1.80:8000/v1/models: 401 Client Error: Unauthorized for url: http://10.244.1.80:8000/v1/models (service_discovery.py:527:vllm_router.service_discovery)
```

This is because the `VLLM_API_KEY` became invalid

3. Force delete a model pod: run `kubectl delete pod POD_NAME --grace-period=0 --force`
4. Monitor the logs of the router

You will see errors in scraping metrics like this:

```
[2025-08-20 10:32:38,681] ERROR: Failed to scrape metrics from http://10.244.1.81:8000: HTTPConnectionPool(host='10.244.1.81', port=8000): Max retries exceeded with url: /metrics (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x735cb0cea6c0>: Failed to establish a new connection: [Errno 113] No route to host')) (engine_stats.py:131:vllm_router.stats.engine_stats)
```

And these errors keeps repeating

If you logged the `self.get_endpoint_info()`, then you will see that the entry corresponding to the deleted pod still appears! This is the `ZombieIP`

# Why are the errors raised

**Main cause**: There is a blocking call within the kubernetes watcher that calls the `/v1/models` endpoint of the model pod. This blocking call can make pause consumption of the Kubernetes watch stream

If during this pause, an event occurred in kubernetes, like the deletion of a pod -> it won't be registered EVER

## Why the ERROR: Failed to get model names is raised and what are the consequences

The Kubernetes watcher watches the kubernetes objects during every 30 seconds, during 30 seconds. There is no `resource_version` (see https://github.com/kubernetes-client/python/blob/master/kubernetes/base/watch/watch.py#L136) set in this watcher, therefore every watch is "new": the watcher thinks the objects detected are all new, therefore the event related to the watch are always `ADDED`, even though the pod was already registered in the `available_engine` object

The logs that prove it are available [here](https://gitlab.com/multiverse1/product/compactifai/compactifai_vllm_helm_chart/-/issues/14#note_2692752307)

When the `/v1/models` starts being slow and raises an error then:
- the function `_get_model_names` returns an empty list -> see [here](https://github.com/vllm-project/production-stack/blob/a4de81ae76187fe4fbd60be1f198864affaffe64/src/vllm_router/service_discovery.py#L522)
- As the event is `ADDED`, then no update is done on the `available_engine` object -> see [here](https://github.com/vllm-project/production-stack/blob/a4de81ae76187fe4fbd60be1f198864affaffe64/src/vllm_router/service_discovery.py#L672)

Therefore, while the pod is up, then we will see a loop of logs `ERROR: Failed to get model names`, AND some traffic will still be routed to this pod as it will stay registered in `available_engine` object -> leading to failure :exploding_head: 

## Why the ERROR: Failed to scrape metrics even when the pod has been deleted
Following the error above, one might think: "I just need to delete the pod and traffic will not be routed to it"

That's the worst thing to do :skull_crossbones: if there are more than 2 faulty pods

The blocking call stops the consumption of the kubernetes watch stream. More simply: if a pod is deleted and completely disappears WHILE the watch stream is blocked, its deletion never gets "watched" by the kubernetes watch stream. The next watch loop will not see that this pod has ever existed and do not exist anymore.

The problem: the `self.available_engine` object is *only* updated within the watches. If a pod is not detected by the watch its corresponding entry in the `self.available_engine` never gets updated.

That is why, if the pod gets deleted while the watch engine is blocked, then its deletion is never registered, therefore the `self.available_engine` is *never* updated and will always hold the entry for the deleted pod

The consequence:
- The stat scraper will keep trying to scrape metrics from the pod: we will see this log `ERROR: Failed to scrape metrics` forever
- Some requests will still be routed to the deleted pod, as the router code uses the `self.available_engine` as source of truth for routing requests -> see [here](https://github.com/vllm-project/production-stack/blob/main/src/vllm_router/services/request_service/request.py#L208)

Where the router routes the request to a pod IP that doesn't exist anymore

### Expected behavior

The `self.available_engine` should stay up to date with the *really* available pods. Some ideas to fix this bug:

- remove blocking call within the k8 watch loop -> make the calls asynchronous, as suggested in https://github.com/vllm-project/production-stack/issues/431
- Pass a ressource version to the k8 watch loop
- Create another mechanism that checks healthiness of `self.available_engine` object outside of the watch loop: check that the pod names present in the `self.available_engine` corresponds to pod that do exist

### Additional context

This is related to https://github.com/vllm-project/production-stack/issues/431

Please feel free to ask for more example and suggestions for fixing the issue 🔥 🚒 🧑‍🚒

zerofishnoodles:
[Bugfix] Fix routing to delete endpoint

- Make k8s watch stream long-lasting connection. 
- Delete event should be handled immediately. 
 

FIX #656  (*link existing issues this PR will resolve*)

---

<details>

<ul>
</ul>

<h3>Code Quality</h3>

<ul>
</ul>

, Shaoting-Feng or ApostaC.

### DIFF
--- a/.github/values-06-session-routing.yaml
+++ b/.github/values-06-session-routing.yaml
@@ -3,11 +3,12 @@ servingEngineSpec:
     type: Recreate
   runtimeClassName: ""
   modelSpec:
-    - name: "opt125m"
+    # Prefill node configuration
+    - name: "opt125m-prefill"
       repository: "lmcache/vllm-openai"
       tag: "2025-05-27-v1"
       modelURL: "facebook/opt-125m"
-      replicaCount: 2
+      replicaCount: 1
       requestCPU: 8
       requestMemory: "30Gi"
       # requestGPU: 1
@@ -18,10 +19,52 @@ servingEngineSpec:
         v1: 1
         gpuMemoryUtilization: 0.6
       lmcacheConfig:
+        cudaVisibleDevices: "0"
         enabled: true
+        kvRole: "kv_producer"
+        enableNixl: true
+        nixlRole: "sender"
+        nixlPeerHost: "vllm-opt125m-decode-engine-service"
+        nixlPeerPort: "55555"
+        nixlBufferSize: "1073741824"  # 1GB
+        nixlBufferDevice: "cuda"
+        nixlEnableGc: true
+        enablePD: true
         cpuOffloadingBufferSize: 0
       labels:
-        model: "opt125m"
+        model: "opt125m-prefill"
+      chatTemplate: "chat.jinja2"
+      chatTemplateConfigMap: |-
+        {% for message in messages %}{{'<|im_start|>' + message['role'] + '\n' + message['content']}}{% if (loop.last and add_generation_prompt) or not loop.last %}{{ '<|im_end|>' + '\n'}}{% endif %}{% endfor %}
+        {% if add_generation_prompt and messages[-1]['role'] != 'assistant' %}{{ '<|im_start|>assistant\n' }}{% endif %}
+    # Decode node configuration
+    - name: "opt125m-decode"
+      repository: "lmcache/vllm-openai"
+      tag: "2025-05-27-v1"
+      modelURL: "facebook/opt-125m"
+      replicaCount: 1
+      requestCPU: 8
+      requestMemory: "30Gi"
+      # requestGPU: 1
+      pvcStorage: "50Gi"
+      vllmConfig:
+        enablePrefixCaching: true
+        maxModelLen: 1024
+        v1: 1
+      lmcacheConfig:
+        cudaVisibleDevices: "1"
+        enabled: true
+        kvRole: "kv_consumer"  # Set decode node as consumer
+        enableNixl: true
+        nixlRole: "receiver"
+        nixlPeerHost: "0.0.0.0"
+        nixlPeerPort: "55555"
+        nixlBufferSize: "1073741824"  # 1GB
+        nixlBufferDevice: "cuda"
+        nixlEnableGc: true
+        enablePD: true
+      labels:
+        model: "opt125m-decode"
       chatTemplate: "chat.jinja2"
       chatTemplateConfigMap: |-
         {% for message in messages %}{{'<|im_start|>' + message['role'] + '\n' + message['content']}}{% if (loop.last and add_generation_prompt) or not loop.last %}{{ '<|im_end|>' + '\n'}}{% endif %}{% endfor %}
--- a/.github/values-06-session-routing.yaml
+++ b/.github/values-06-session-routing.yaml
@@ -3,12 +3,11 @@ servingEngineSpec:
     type: Recreate
   runtimeClassName: ""
   modelSpec:
-    # Prefill node configuration
-    - name: "opt125m-prefill"
+    - name: "opt125m"
       repository: "lmcache/vllm-openai"
       tag: "2025-05-27-v1"
       modelURL: "facebook/opt-125m"
-      replicaCount: 1
+      replicaCount: 2
       requestCPU: 8
       requestMemory: "30Gi"
       # requestGPU: 1
@@ -19,52 +18,10 @@ servingEngineSpec:
         v1: 1
         gpuMemoryUtilization: 0.6
       lmcacheConfig:
-        cudaVisibleDevices: "0"
         enabled: true
-        kvRole: "kv_producer"
-        enableNixl: true
-        nixlRole: "sender"
-        nixlPeerHost: "vllm-opt125m-decode-engine-service"
-        nixlPeerPort: "55555"
-        nixlBufferSize: "1073741824"  # 1GB
-        nixlBufferDevice: "cuda"
-        nixlEnableGc: true
-        enablePD: true
         cpuOffloadingBufferSize: 0
       labels:
-        model: "opt125m-prefill"
-      chatTemplate: "chat.jinja2"
-      chatTemplateConfigMap: |-
-        {% for message in messages %}{{'<|im_start|>' + message['role'] + '\n' + message['content']}}{% if (loop.last and add_generation_prompt) or not loop.last %}{{ '<|im_end|>' + '\n'}}{% endif %}{% endfor %}
-        {% if add_generation_prompt and messages[-1]['role'] != 'assistant' %}{{ '<|im_start|>assistant\n' }}{% endif %}
-    # Decode node configuration
-    - name: "opt125m-decode"
-      repository: "lmcache/vllm-openai"
-      tag: "2025-05-27-v1"
-      modelURL: "facebook/opt-125m"
-      replicaCount: 1
-      requestCPU: 8
-      requestMemory: "30Gi"
-      # requestGPU: 1
-      pvcStorage: "50Gi"
-      vllmConfig:
-        enablePrefixCaching: true
-        maxModelLen: 1024
-        v1: 1
-      lmcacheConfig:
-        cudaVisibleDevices: "1"
-        enabled: true
-        kvRole: "kv_consumer"  # Set decode node as consumer
-        enableNixl: true
-        nixlRole: "receiver"
-        nixlPeerHost: "0.0.0.0"
-        nixlPeerPort: "55555"
-        nixlBufferSize: "1073741824"  # 1GB
-        nixlBufferDevice: "cuda"
-        nixlEnableGc: true
-        enablePD: true
-      labels:
-        model: "opt125m-decode"
+        model: "opt125m"
       chatTemplate: "chat.jinja2"
       chatTemplateConfigMap: |-
         {% for message in messages %}{{'<|im_start|>' + message['role'] + '\n' + message['content']}}{% if (loop.last and add_generation_prompt) or not loop.last %}{{ '<|im_end|>' + '\n'}}{% endif %}{% endfor %}
--- a/.github/values-07-prefix-routing.yaml
+++ b/.github/values-07-prefix-routing.yaml
@@ -3,11 +3,12 @@ servingEngineSpec:
     type: Recreate
   runtimeClassName: ""
   modelSpec:
-    - name: "opt125m"
+    # Prefill node configuration
+    - name: "opt125m-prefill"
       repository: "lmcache/vllm-openai"
       tag: "2025-05-27-v1"
       modelURL: "facebook/opt-125m"
-      replicaCount: 2
+      replicaCount: 1
       requestCPU: 8
       requestMemory: "30Gi"
       # requestGPU: 1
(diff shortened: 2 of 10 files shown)


jonoillar:
I'll test tmr further if your fix works for my specific use case.

While this might resolve it, there is one use case that becomes problematic with your quick fix:
Situation: 2 model pods (`pod_1` and `pod_2`)
- `pod_1` gets an event ADDED associated to it -> it triggers a blocking `/v1/models` endpoint. The consumption of the event loop is blocked, i.e. the watcher doesn't watch
- While the watcher doesn't watch, `pod_2` is deleted -> it's associated entry in `self.available_engine` won't be updated until the watcher resumes watching

For the entire time the watcher is blocked, the request directed to `pod_2` will fail.

This is why I think we need to decouple the watch loop and handling the events, which is the purpose of this MR: https://github.com/vllm-project/production-stack/pull/666. And handling the events in an asynchronous manner

Furthermore, I think that this MR still allows an unexperienced user to set `k8s-watcher-timeout-seconds>0` and experience the issue I've been experiencing initially in https://github.com/vllm-project/production-stack/issues/656

ruizhang0101:
> I'll test tmr further if your fix works for my specific use case.
> 
> While this might resolve it, there is one use case that becomes problematic with your quick fix: Situation: 2 model pods (`pod_1` and `pod_2`)
> 
> * `pod_1` gets an event ADDED associated to it -> it triggers a blocking `/v1/models` endpoint. The consumption of the event loop is blocked, i.e. the watcher doesn't watch
> * While the watcher doesn't watch, `pod_2` is deleted -> it's associated entry in `self.available_engine` won't be updated until the watcher resumes watching
> 
> For the entire time the watcher is blocked, the request directed to `pod_2` will fail.
> 
> This is why I think we need to decouple the watch loop and handling the events, which is the purpose of this MR: #666. And handling the events in an asynchronous manner
> 
> Furthermore, I think that this MR still allows an unexperienced user to set `k8s-watcher-timeout-seconds>0` and experience the issue I've been experiencing initially in #656

So from another perspective, can we just handle v1/models so that when it is way longer than expected, we just abandon it and continue the loop? because the v1/models is supposed to be fast and "non-blocking". 

And I do believe that using a consumer process will make the design more efficient, but I think we may need more test for the big change.

jonoillar:
> So from another perspective, can we just handle v1/models so that when it is way longer than expected, we just abandon it and continue the loop? because the v1/models is supposed to be fast and "non-blocking".

We could, but I don't think we should assume anything about how fast the `v1/models` answers. I think we should just be prepared to any case (backend being fast and backend being slow). From my usage, I've been facing backends that displayed this exact behaviour, due sometimes to connectivity slowdowns or other.

In the end, I don't think we should assume anything about the healthiness backend: it can be fast or slow, some endpoint can be fast or slow, there could be connectivity errors. But the vllm router should be agnostic to this and be prepared to handle different backends

ruizhang0101:
> > So from another perspective, can we just handle v1/models so that when it is way longer than expected, we just abandon it and continue the loop? because the v1/models is supposed to be fast and "non-blocking".
> 
> We could, but I don't think we should assume anything about how fast the `v1/models` answers. I think we should just be prepared to any case (backend being fast and backend being slow). From my usage, I've been facing backends that displayed this exact behaviour, due sometimes to connectivity slowdowns or other.
> 
> In the end, I don't think we should assume anything about the healthiness backend: it can be fast or slow, some endpoint can be fast or slow, there could be connectivity errors. But the vllm router should be agnostic to this and be prepared to handle different backends

I understand it and I think this statement is legit, but I am still a little confused about the event consumer process, if the endpoint is slow, it will block the event consumer process and not delete the endpoint in time in the same manner as you said before.

jonoillar:
> if the endpoint is slow, it will block the event consumer process and not delete the endpoint in time in the same manner as you said before.

True, it will block the event consumer thread if the call to the endpoint is done within this thread. Hence making this call in another thread (by decoupling event consumer and event handler) *asynchronously* would do:
- call the blocking endpoint for `pod_1`
- event loop processes another event in the meantime -> successfully deleted `pod_2`
- event loop goes back checking if the blocking endpoint for `pod_1` has finished

I have to say I'm a bit confused about your statement though, could you explain your point a bit more in detail? Thanks !! 🙏 

By the way, the kubernetes watcher is synchronous by default. There is an asynchronous watcher available in https://github.com/tomplus/kubernetes_asyncio , but I didn't want to use it since I don't think it receives that much support as the official kubernetes python api 😄

ruizhang0101:
> > if the endpoint is slow, it will block the event consumer process and not delete the endpoint in time in the same manner as you said before.
> 
> True, it will block the event consumer thread if the call to the endpoint is done within this thread. Hence making this call in another thread (by decoupling event consumer and event handler) _asynchronously_ would do:
> 
> * call the blocking endpoint for `pod_1`
> * event loop processes another event in the meantime -> successfully deleted `pod_2`
> * event loop goes back checking if the blocking endpoint for `pod_1` has finished
> 
> I have to say I'm a bit confused about your statement though, could you explain your point a bit more in detail? Thanks !! 🙏
> 
> By the way, the kubernetes watcher is synchronous by default. There is an asynchronous watcher available in https://github.com/tomplus/kubernetes_asyncio , but I didn't want to use it since I don't think it receives that much support as the official kubernetes python api 😄

I see what you mean. Could you do some test and stress test to make sure the consistency for the shared resource like available_engines? 

Yeah I don't think the asyncio k8s will be a good choice at this moment as well lolll.

jonoillar:
> I see what you mean. Could you do some test and stress test to make sure the consistency for the shared resource like available_engines?

For my PR https://github.com/vllm-project/production-stack/pull/666 ? Yep I can do it, do you have specific test and stress test scenarios in mind ?

ruizhang0101:
> > I see what you mean. Could you do some test and stress test to make sure the consistency for the shared resource like available_engines?
> 
> For my PR #666 ? Yep I can do it, do you have specific test and stress test scenarios in mind ?

Yep, for PR #666 . 

There is a simple stress test script under the tests/e2e, but it is request level not cluster operations
Other than that, we might need to imitate some scenarios including multiple operations like add pod/delete pods in different order.

YuhanLiu11:
Hey @jonoillar, thanks for submitting PR #666 to address the zombie IP issue in the production stack. As discussed in this PR, I think this change can be a temporary fix for #656 with minimal changes on the current codebase. For this reason, I’m merging this PR first as a quick fix for issue #656.

That said, I’ll be happy to review PR #666 once you have some testing numbers and have verified that it doesn’t break any functionalities in the router.

YuhanLiu11 (approved):

jonoillar:
@YuhanLiu11 @zerofishnoodles I tested your MR and it does fix the issue I was facing.

Now it also created 2 problematic scenarios:

# Scenario 1: unhealthy pod removed forever

Set up:
- 1 router pod
- 1 model pod
- Add logic within router so that the `/v1/models` endpoint fails at the 2nd iteration

What happens:
1. After a round of ADDED - MODIFIED MODIFIED MODIFIED ... , the pod finally gets added in the service discovery -> the /`v1/models` has been called once
2. Here do a "dummy" modif to the pod -> for example, add a label to the pod
   - a new MODIFIED event related to the pod will be triggered. This event will trigger a faulty  `/v1/models` endpoint
   - The endpoint fails -> the pod gets deleted

And the pod will be never again added to the available engine. 

**REAL LIFE SCENARIO**: with an instable backend that has this characteristics:
- /v1/models fails a few times (for any reason - connectivity, pod freeze, etc)
- the pod never changes so that there is never a MODIFIED event triggered related to the pod

With this setup, then the first time the pod is deleted, it will *never* be discovered again

**DAMAGE**: service discovery won't discover a pod. The pod will use up GPU's for nothing, and other pod will suffer higher load

# Scenario 2: Pod deleted, some request still routed to it
Set up:
- 1 router pod
- 1 old pod
- 9 new pods with blocking `v1/models` call

What happens:
`events_queue = []`
- The first 9 pods are created and ready at the same time. There are 9 "ADDED" event that will lead to a call to `v1/models` lined up ->`events_queue = ["ADDED", "ADDED", ..., "ADDED"]`
- The old pod got deleted: one event DELETED is added to the list of events to process: `events_queue = ["ADDED", "ADDED", ..., "ADDED", "DELETED"]`

Each of the "ADDED" event takes `self.health_check_timeout_seconds=10s` to be processed. So in total `9*self.health_check_timeout_seconds=90s` until the "DELETED" event is processed

During this time, some request will be routed to the old pod -> they will all fail

**DAMAGE**: Within a window of time, request will be routed to deleted pod. And this window of time can be proportional to the number of pods to be served

**POSSIBLE SOLUTION**: the events related to the pods are tackled successively, whereas they could be tackled concurrently. The events related to `pod_1` should not wait for the event related to `pod_2` to be processed, before being processed.

# Conclusion

Scenario 1 is quite likely to happen and is harmful, as one model pod will be _silently_ ignored. 

Scenario 2 is more likely to happen, with an impact (a window within which request are routed to deleted pod) that is varying. I'd say for most of the time this won't be critical, but if this code is deployed in a prod environment were very high SLA (Availability) are required, this can matter a lot


Please tell me what are your thoughts !! If I'm not clear, please ask 😄

PR #675, opened 2025-09-03
[Misc] Add Makefile to run full pipeline with a single command

ikaadil:
### Summary
This PR introduces a Makefile that simplifies running the entire pipeline with a single command.  
It aims to improve developer experience by reducing the need to remember or manually chain multiple commands.

### Changes
- Added a `Makefile` with targets to run the full pipeline
- Ensures consistency across environments and local development workflows

### Benefits
- Provides a single entry point for running the pipeline
- Improves developer productivity and onboarding
- Helps standardize the workflow for contributors

### Notes
- No breaking changes introduced
- Passed all linter checks and GitHub workflows locally

### DIFF
--- a/observability/README.md
+++ b/observability/README.md
@@ -19,7 +19,7 @@ Make sure to have:
 After that you can run:
 
 ```bash
-sudo bash install.sh
+bash install.sh
 ```
 
 After installing, the dashboard can be accessed through the service `service/kube-prom-stack-grafana` in the `monitoring` namespace.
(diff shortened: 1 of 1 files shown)


YuhanLiu11:
@zerofishnoodles Can you review this PR? Thanks

zerofishnoodles (commented):
I think these kind of helpers are good but are not supposed to be in the repo but just a personal helper, at least it should not be in the main directory as Makefile but rather in a helper dir

ikaadil:
> I think these kind of helpers are good but are not supposed to be in the repo but just a personal helper, at least it should not be in the main directory as Makefile but rather in a helper dir

 Thanks for the feedback! 🙏 

The main reason I added the `Makefile` at the repo root is to make common tasks (tests, linting, setup, etc.) easier for all contributors—so nobody has to remember long commands. This is a fairly common practice in Python projects (similar to keeping `requirements.txt` or `pyproject.toml` at the root).

I’m open to moving it under `helpers/` or `scripts/` if the team prefers. I just think having it at the root lowers the barrier for new contributors and keeps workflows consistent.

PR #681, opened 2025-09-04
[Bugfix][Router]: loop through model_names

max-wittig:
Fixes this crash. `model_name` does not exist:

```sh
[2025-09-04 15:47:55,962] DEBUG: ==== Enter audio_transcriptions ==== (request.py:542:vllm_router.services.request_service.request)
[2025-09-04 15:47:55,962] DEBUG: Received upload: example.m4a (audio/mp4) (request.py:543:vllm_router.services.request_service.request)
[2025-09-04 15:47:55,962] DEBUG: Params: model=whisper-large-v3-turbo prompt=None response_format='json' temperature=None language=en (request.py:544:vllm_router.services.request_service.request)
[2025-09-04 15:47:55,963] DEBUG: ==== Total endpoints ==== (request.py:568:vllm_router.services.request_service.request)
[...]

PR #705, opened 2025-09-19
[Bugfix][Router] use model_type instead of model_label

max-wittig:
As model_label is already used for
things related to the LMCache. It
makes more sense to use the type of
the model here.

Replaces https://github.com/vllm-project/production-stack/pull/681

---

- [ ] Make sure the code changes pass the [pre-commit](https://github.com/vllm-project/production-stack/blob/main/CONTRIBUTING.md) checks.

<details>

<ul>
</ul>

<h3>Code Quality</h3>

<ul>
</ul>

, Shaoting-Feng or ApostaC.

### DIFF
--- a/src/vllm_router/services/request_service/request.py
+++ b/src/vllm_router/services/request_service/request.py
@@ -585,13 +585,13 @@ async def route_general_transcriptions(
 
     endpoints = service_discovery.get_endpoint_info()
 
-    # filter the endpoints url by model name and model label for transcriptions
+    # filter the endpoints url by model name and model type for transcriptions
     transcription_endpoints = []
     for ep in endpoints:
         for model_name in ep.model_names:
             if (
                 model == model_name
-                and ep.model_label == "transcription"
+                and ep.model_type == "transcription"
                 and not ep.sleep
             ):
                 transcription_endpoints.append(ep)
--- a/src/vllm_router/service_discovery.py
+++ b/src/vllm_router/service_discovery.py
@@ -97,6 +97,9 @@ class EndpointInfo:
     # Endpoint's sleep status
     sleep: bool
 
+    # Model type (e.g., "transcription", "chat", etc.)
+    model_type: Optional[str] = None
+
     # Pod name
     pod_name: Optional[str] = None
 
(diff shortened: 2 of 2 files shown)


max-wittig:
Let me know, what you think about this @zerofishnoodles

src/vllm_router/services/request_service/request.py
```diff
@@ -585,13 +585,13 @@ async def route_general_transcriptions(
...
 
-    # filter the endpoints url by model name and model label for transcriptions
+    # filter the endpoints url by model name and model type for transcriptions
     transcription_endpoints = []
     for ep in endpoints:
         for model_name in ep.model_names:
             if (
                 model == model_name
-                and ep.model_label == "transcription"
+                and ep.model_type == "transcription"
```
ruizhang0101, line 594: Do we actually need model type here? Wouldn't model name itself be enough since we know which model is for transcription when querying.

max-wittig: @zerofishnoodles Not sure I can follow here. How do we know which model name is capable of `transcription`?

E.g. we're using `whisper-large-v3-turbo`, but there are others.

ruizhang0101: I mean when we query we would specify the model name which means we have already known the model is capable of something. You wouldn't query llama for transcription. Unless we want one model to serve specific task, in that case If we really want to add another annotation for that, I would say model_task is more suitable since transcription is not a model type.

max-wittig: ahhh you're right. I made https://github.com/vllm-project/production-stack/pull/712 so that we can focus on that. Do we still want the model types for the other Kubernetes based discovery modes?

We're not using it, but I guess it would be useful. Right now the real healthcheck just works for the static service discovery.

ruizhang0101: Yeah I guess we can do that and it will be helpful. Currently we only use /health to inspect the health. If we want to do that, I think we need to add some readme or tutorials so that people know they need to have label on it. Speaking of that, we may also need to modify the helm chart since pod level label is not supported right now.

PR #711, opened 2025-09-22
[Feat][Router] Add HashTrie LRU to prevent OOM

### LINKED ISSUE #687: bug: router crashes because of OOM and prefix cache HashTrie memory usage is unbounded
### Describe the bug

We found an issue during the stability test that router Pod keeps restarting after running for a certain amount of time. It seems the router is restarted by k8s because of OOM. We configured the router Pod limit with 2G memory initially. After raising to 16G, we observed that the memory usage of router is unbounded and keeps ramping up.

### To Reproduce

1. Create a vLLM production stack router, connected with multiple vLLM model server Pods
2. Configure the memory limit of router to be 2GB
3. Run a continuous long document QA benchmark for over 2 hours
4. The router pod should be periodically get restarted because of OOM

### Expected behavior

The HashTrie size should be capped by a limit and perhaps a GC can run periodically to cleanup the stale nodes to free up the memory, e.g TTL or LRU

### Additional context

Does vLLM production stack team have any plan to fix this?

can-sun:
### Description
Currently the prefix aware routing HashTrie misses the capability of node eviction. This will result in router pod crashes after the memory is filled up with TrieNodes. During pod restart period, users will experience downtime and in-progress requests could be terminated in the middle stream. 

This PR adds LRU for HashTrie to evict the nodes when hitting memory limit. It will resolve the issue that router keeps restarting by Kubernetes due to insufficient memory. Besides, HashTrieConfig is added to configure some important parameters that will affect the performance, e.g eviction threshold, max memory limit, target utilization etc. These are configured via environment variables.

Nodes are added to the LRU when they are inserted to the Trie. When this node is revisited by other request, the node will be popped and added to tail of LRU. This proposed approach will evict the child nodes first since child nodes are always less frequently accessed compared to parent.

### Memory Usage before addubg Eviction
The periodic up and downs are router pod restarts because of OOM. After allocating sufficient memory (over 16GB), the memory usage goes unbounded.

image

### Memory Usage After adding Eviction
After reaching the eviction threshold, nodes will start being evicted to target utilization. Thus we identify that memory usage keeps stable without ramping up anymore.

image

FIX #687 (*link existing issues this PR will resolve*)

---

<details>

<ul>
</ul>

<h3>Code Quality</h3>

<ul>
</ul>

, Shaoting-Feng or ApostaC.

### DIFF
--- a/src/vllm_router/prefix/config.py
+++ b/src/vllm_router/prefix/config.py
@@ -1,0 +1,58 @@
+# Copyright 2024-2025 The vLLM Production Stack Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+from dataclasses import dataclass
+from typing import Optional
+
+
+@dataclass
+class HashTrieConfig:
+    """Configuration for HashTrie with LRU eviction"""
+
+    chunk_size: int = 0
+    max_memory_size: int = 0  # GB
+    eviction_threshold: float = 0  # start evicting at
+    target_utilization: float = 0  # evict down to
+    memory_check_request_batch_size: int = 0
+
+    @staticmethod
+    def from_defaults(
+        chunk_size: int = 128,
+        max_memory_size: int = 2,
+        eviction_threshold: float = 0.9,
+        target_utilization: float = 0.5,
+        memory_check_request_batch_size: int = 10,
+    ) -> "HashTrieConfig":
+        """Create configuration with default values"""
+        return HashTrieConfig(
+            chunk_size,
+            max_memory_size,
+            eviction_threshold,
+            target_utilization,
+            memory_check_request_batch_size,
+        )
+
+    @staticmethod
+    def from_env() -> "HashTrieConfig":
+        """Load configuration from environment variables"""
+        return HashTrieConfig(
+            chunk_size=int(os.getenv("HASHTRIE_CHUNK_SIZE", "128")),
+            max_memory_size=int(os.getenv("PREFIXAWARE_MAX_MEMORY_SIZE_GB", "2")),
+            eviction_threshold=float(os.getenv("HASHTRIE_EVICTION_THRESHOLD", "0.9")),
+            target_utilization=float(os.getenv("HASHTRIE_TARGET_UTILIZATION", "0.5")),
+            memory_check_request_batch_size=int(
+                os.getenv("MEMORY_CHECK_REQUEST_BATCH_SIZE", "10")
+            ),
+        )
--- a/src/vllm_router/prefix/hashtrie.py
+++ b/src/vllm_router/prefix/hashtrie.py
@@ -14,17 +14,30 @@
 
 import asyncio
 import logging
-from typing import Generator, Set, Tuple
+import os
+from collections import OrderedDict
+from typing import Dict, Generator, List, Optional, Set, Tuple
 
+import psutil
 import xxhash
+
+from .config import HashTrieConfig
 
 logger = logging.getLogger(__name__)
 
 
+class NodeMetadata:
+    """Metadata for nodes tracked in LRU cache"""
+
+    def __init__(self, parent: "TrieNode", child_hash: int):
+        self.parent = parent
+        self.child_hash = child_hash
+
+
 class TrieNode:
     def __init__(self):
-        self.children = {}
-        self.endpoints = set()
+        self.children: Dict[int, "TrieNode"] = {}
+        self.endpoints: Set[str] = set()
 
         # assign a lock for each trie node.
         # this assures that each node will only be accessed by one co-routine
@@ -33,14 +46,27 @@ class TrieNode:
 
 
 class HashTrie:
-    def __init__(self, chunk_size: int = 128):
+    def __init__(self, config: Optional[HashTrieConfig] = None):
         """
-        Initialize the HashTrie.
+        Initialize the HashTrie with LRU eviction based on system memory pressure.
         Args:
-            chunk_size (int): the string chunk size (in terms of # characters)
+            config (HashTrieConfig): Configuration for the HashTrie
         """
+        self.config = config or HashTrieConfig.from_defaults()
         self.root = TrieNode()
-        self.chunk_size = chunk_size
+
+        # HashTrie LRU
+        self.node_cache = OrderedDict[TrieNode, NodeMetadata]()
+        # eviction threshold
+        self.memory_threshold_mb = (
+            self.config.max_memory_size * 1024 * self.config.eviction_threshold
+        )
+        # eviction percentage
+        self.eviction_target_percentage = 1.0 - self.config.target_utilization
+        self.current_memory_mb = 0.0
+        self.operation_count = 0
+        # ensure that eviction is performed by only one co-routine
+        self.eviction_lock = asyncio.Lock()
 
     def _chunk_and_hash(self, request: str) -> Generator[int, None, None]:
         """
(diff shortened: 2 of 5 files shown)


YuhanLiu11:
@can-sun Thanks for the fix! Could you fix the pre-commit error and show an easy way for us to see the effect of your fix?

can-sun:
> @can-sun Thanks for the fix! Could you fix the pre-commit error and show an easy way for us to see the effect of your fix?

Thanks Yuhan, let me double check the pre-commit failures tmr. I am pretty sure I have run the pre-commit check before sending out the PR. I will pull up some metrics that I can share with you tomorrow to justify the effect.

YuhanLiu11:
@can-sun Thanks for the fix. BTW I realize that that if we evict nodes from the HashTrie, the vLLM-side delay of the requests may be affected as the information of where did the requests go is missing. We will have to run some systematic benchmarks to test the e2e performance of your PR. 
cc @zerofishnoodles

can-sun:
> @can-sun Thanks for the fix. BTW I realize that that if we evict nodes from the HashTrie, the vLLM-side delay of the requests may be affected as the information of where did the requests go is missing. We will have to run some systematic benchmarks to test the e2e performance of your PR. cc @zerofishnoodles

You are definitely right. We also did some benchmarking locally and evaluated the performance impact. Let me share that with you today.

jiapinw:
> the vLLM-side delay of the requests may be affected as the information of where did the requests go is missing

I think this is intentional for this PR. It's a balance between availability and latency imo. There won't be any performance regression until we hit the configured mem limit. If I am reading the PR correctly, the node eviction will only happen when we reach the limit, ie. if we trim down a root node, then it will free up quite a few space and it would give an extended period of time when eviction is no longer needed. Though we do need to be careful regarding how we set the default, one improvement area might be setting the default based on instance type or other env var, for instance cpu limits configured during a k8s deployment.

can-sun:
Done, added some data collected during benchmarking. Hope it helps. Feel free to take in the code and do the local performance evaluations. Thanks

can-sun:
Hey @YuhanLiu11 Do you know and next steps for this change? Currently we are waiting for this change to be in place before putting the router into use.

can-sun:
Double checked the E2E test failed because `ERROR: Could not install packages due to an OSError: [Errno 28] No space left on device`

YuhanLiu11:
hey @can-sun thanks for adding the benchmarking results! 

@zerofishnoodles can you check why this CI fails due to space errors? Is our CI server full?

ruizhang0101:
This is weird. We just cleaned the space not long ago. I will take a look today.

ruizhang0101:
That one is the GitHub hosted machine, their machine is out of space. :(
I think from our side, we can try to modify the CI to either reduce the dependency size or move it to our self-hosted machine

src/vllm_router/prefix/config.py
```diff
@@ -0,0 +1,57 @@
...
+            target_utilization,
+            memory_check_request_batch_size,
+        )
+
+    @staticmethod
+    def from_env() -> "HashTrieConfig":
+        """Load configuration from environment variables"""
+        return HashTrieConfig(
+            chunk_size=int(os.getenv("HASHTRIE_CHUNK_SIZE", "128")),
+            max_memory_size=int(os.getenv("PREFIXAWARE_MAX_MEMORY_SIZE_GB", "2")),
```
piyushdaftary, line 51: Read default values from constant instead of hardcoding here

src/vllm_router/prefix/hashtrie.py
```diff
@@ -14,17 +14,30 @@
...
+import psutil
 import xxhash
 
+from .config import HashTrieConfig
+
 logger = logging.getLogger(__name__)
 
 
+class NodeMetadata:
+    """Metadata for nodes tracked in LRU cache"""
```
piyushdaftary, line 30: You need to make NodeMetadata to be immutable for nodes tracked in LRU cache. Please add `@dataclass(frozen=True)`

src/vllm_router/prefix/hashtrie.py
```diff
@@ -14,17 +14,30 @@
...
+        self.parent = parent
+        self.child_hash = child_hash
+
+
 class TrieNode:
     def __init__(self):
-        self.children = {}
-        self.endpoints = set()
+        self.children: Dict[int, "TrieNode"] = {}
+        self.endpoints: Set[str] = set()
```
piyushdaftary, line 40: Suggest to add is_Leaf and other operation methods, something as below

```
    @property
    def is_leaf(self) -> bool:
        """Check if the node is a leaf node."""
        return not bool(self.children)

    async def add_child(self, hash_value: int, node: "TrieNode") -> None:
        """Add a child node."""
        async with self.lock:
            self.children[hash_value] = node

    async def remove_child(self, hash_value: int) -> None:
        """Remove a child node."""
        async with self.lock:
            self.children.pop(hash_value, None)

    async def add_endpoint(self, endpoint: str) -> None:
        """Add an endpoint to the node."""
        async with self.lock:
            self.endpoints.add(endpoint)

    def __repr__(self) -> str:
        return f"TrieNode(children={len(self.children)}, endpoints={len(self.endpoints)})"
```

src/vllm_router/prefix/hashtrie.py
```diff
@@ -33,14 +46,27 @@ def __init__(self):
...
-        self.chunk_size = chunk_size
+
+        # HashTrie LRU
+        self.node_cache = OrderedDict[TrieNode, NodeMetadata]()
+        # eviction threshold
+        self.memory_threshold_mb = (
+            self.config.max_memory_size * 1024 * self.config.eviction_threshold
+        )
+        # eviction percentage
+        self.eviction_target_percentage = 1.0 - self.config.target_utilization
```
piyushdaftary, line 65: Please move all these constant values to be read from env variables or default constant . Also see it it make sense to introduce HashTrieStats something like

```
@dataclass
class HashTrieStats:
    current_memory_mb: float = 0.0
    operation_count: int = 0

@dataclass
class HashTrie:
    config: HashTrieConfig = field(default_factory=HashTrieConfig.from_defaults)
    root: TrieNode = field(default_factory=TrieNode)
    node_cache: OrderedDict[TrieNode, NodeMetadata] = field(default_factory=OrderedDict)
    stats: HashTrieStats = field(default_factory=HashTrieStats)
    eviction_lock: asyncio.Lock = field(default_factory=asyncio.Lock)

    def __post_init__(self):
        self.memory_threshold_mb = (
            self.config.max_memory_size * BYTES_TO_MB * self.config.eviction_threshold
        )
        self.eviction_target_percentage = 1.0 - self.config.target_utilization

    @property
    def current_memory_mb(self) -> float:
        return self.stats.current_memory_mb

    @current_memory_mb.setter
    def current_memory_mb(self, value: float) -> None:
        self.stats.current_memory_mb = value

    @property
    def operation_count(self) -> int:
        return self.stats.operation_count

    @operation_count.setter
    def operation_count(self, value: int) -> None:
        self.stats.operation_count = value

    def __repr__(self) -> str:
        return (f"HashTrie(nodes={len(self.node_cache)}, "
                f"memory={self.current_memory_mb:.2f}MB)")
```

src/vllm_router/prefix/hashtrie.py
```diff
@@ -51,28 +77,74 @@ def _chunk_and_hash(self, request: str) -> Generator[int, None, None]:
...
+
+        async with self.eviction_lock:
+            # Track operations and check if eviction is needed
+            self.operation_count = (
+                self.operation_count + 1
+            ) % self.config.memory_check_request_batch_size
+
+            # Check if we should evict
+            if self.operation_count == 0:
+                try:
```
piyushdaftary, line 130: can you please move the eviction logic to a separate method in HashTrie, something like

```
class HashTrie:
    async def _check_and_perform_eviction(self) -> None:
        """Check memory usage and perform eviction if necessary."""
        try:
            stats = await MemoryMonitor.get_memory_stats(self.memory_threshold_mb)

            if abs(stats.current_mb - self.current_memory_mb) < 0.1:
                logger.debug(f"Eviction skipped - minimal memory change: {stats.current_mb:.1f}MB")
                return

            self.current_memory_mb = stats.current_mb

            if stats.current_mb > stats.threshold_mb:
                logger.warning(f"Memory threshold exceeded: {stats.current_mb:.1f}MB > {stats.threshold_mb:.1f}MB")
                await self._perform_eviction()
         except Exception as e: 
           // Add Exception Handling 


    async def _perform_eviction(self) -> None:
        """Handle the actual eviction process with retry logic."""
        max_retries = self.config.eviction_max_retries
        for retry in range(max_retries):
            try:
                await self.batch_evict(self.eviction_target_percentage)
                return
            except Exception as e:
           // Add Exception Handling 


        logger.error("Max eviction retries reached")

    async def _handle_monitoring_failure(self) -> None:
        """Handle cases where memory monitoring fails."""
```

PR #733, opened 2025-10-13
fix: Exclude content-length from response headers in route_general_transcriptions

fidoriel:
This ignores the content-length for transcription endpoint because the content length gets wrong sometimes if the encoding by router differs from encoding by server (which can be handled by the code, just clients go crazy if they expect more data)


fidoriel:
@zerofishnoodles the broken ci looks like the k8s setup is broken and not related to this pr ^^

ruizhang0101:
> @zerofishnoodles the broken ci looks like the k8s setup is broken and not related to this pr ^^

Yes! The CI was broken. Fixed, rerunning the ci now :)

bennorris123:
Hi guys, 

Noticed an issue with /v1/audio/transcriptions: the router doesn’t forward client headers Iie: Authorization) to the backend. This causes 401 Unauthorized responses when the backend requires an auth key.

In route_general_transcriptions, the proxy call is:

backend_response = await client.post(
    f"{chosen_url}{endpoint}",
    data=form_data,
)

No headers are passed, which forwards all relevant headers. Note that the chat completions endpoints works fine, meaning route_general_requests works okay.

Fix would be to forward the incoming request headers (similar to process_request) so transcription requests do not result in 401 errors when using the vllm api key in requests.

Many thanks

ruizhang0101:
> Hi guys,
> 
> Noticed an issue with /v1/audio/transcriptions: the router doesn’t forward client headers Iie: Authorization) to the backend. This causes 401 Unauthorized responses when the backend requires an auth key.
> 
> In route_general_transcriptions, the proxy call is:
> 
> backend_response = await client.post( f"{chosen_url}{endpoint}", data=form_data, )
> 
> No headers are passed, which forwards all relevant headers. Note that the chat completions endpoints works fine, meaning route_general_requests works okay.
> 
> Fix would be to forward the incoming request headers (similar to process_request) so transcription requests do not result in 401 errors when using the vllm api key in requests.
> 
> Many thanks

Thank you for bringing this up, it is a good point!  @fidoriel can you also take care of this bug in this pr too since it is relevant to the transcription header.

bennorris123:
Thanks @zerofishnoodles - @fidoriel let me know if you need any more support or content for this issue

bennorris123:
Also @zerofishnoodles @fidoriel do you anticipate this issue being resolved and released this week?

ruizhang0101:
> Also @zerofishnoodles @fidoriel do you anticipate this issue being resolved and released this week?

As long as the PR is merged, it will publish the new docker image by CI :)))

PR #737, opened 2025-10-16
[Bugfix] Allow single-label hostnames in URL validation

osaimi:
Fix validate_url to accept URLs with single-label hosts like http://oss-20b:8000, which are common in Docker and internal networking setups.


ruizhang0101:
Hi, can you fix the gemini review?

PR #768, opened 2025-12-15
[Bugfix][Router] Fix router startup race when using multiple replicas

bcdonadio:
## Summary

This PR fixes a non-deterministic startup race in the router that becomes very easy to trigger when running **two or more router replicas** with Kubernetes service discovery enabled.

In affected deployments, one of the router pods can log an initialization error similar to:

```text
ERROR: Error initializing client sessions: 'State' object has no attribute 'event_loop' (service_discovery.py:714:vllm_router.service_discovery)
```

The fix introduces explicit coordination between the FastAPI event loop and the Kubernetes watcher thread so that background threads only schedule async work after the loop exists.

## Problem

The router uses a background watcher thread to discover engine pods when `serviceDiscovery=k8s` and `k8sServiceDiscoveryType=pod-ip`.

During startup, the watcher can discover engines and attempt to (re)initialize aiohttp client sessions for prefill/decode routing.
When the router is starting up at the same time, the FastAPI event loop is not yet attached to `app.state`, which results in:

- `AttributeError: 'State' object has no attribute 'event_loop'`
- A `RuntimeWarning` about an un-awaited coroutine (because the error occurs while evaluating the loop argument)

This is noisy and can leave the router without the intended client-session initialization at the moment it starts serving.

## Root cause

Two things were happening at the same time:

1. `K8sPodIPServiceDiscovery.__init__()` starts the Kubernetes watcher thread immediately.
2. `app.state.event_loop` was only set later during FastAPI lifespan startup.

If the watcher thread sees an engine pod before `app.state.event_loop` exists, `_add_engine()` attempts to do:

```python
asyncio.run_coroutine_threadsafe(self.initialize_client_sessions(), self.app.state.event_loop)
```

and fails with the missing `event_loop` attribute.

## Why this was mostly showing up with 2 replicas or more

With a single router replica, the router often starts before the engine becomes Ready.
In that typical timeline, the FastAPI lifespan completes startup, the event loop is available, and only then does the watcher observe ready engine pods.

When running 2+ router replicas, the odds of hitting the bad timeline rise sharply:

- The second router pod often starts while the engine is already Ready.
- Its watcher thread receives an immediate ADDED event for the already-running engine.
- That discovery can happen before FastAPI lifespan attaches the event loop, which triggers the error.

In short: multi-replica router startup increases the likelihood that a router instance will observe existing engines during its own initialization window.

## Fix

This PR makes event-loop availability explicit and removes the race:

1. **Attach the running event loop early in lifespan** (`src/vllm_router/app.py`)
   - Set `app.state.event_loop` at the start of the lifespan context, before any async initialization that depends on it.
2. **Add an event-loop handshake to K8s pod-IP service discovery** (`src/vllm_router/service_discovery.py`)
   - `K8sPodIPServiceDiscovery` now exposes `set_event_loop(loop)`.
   - The watcher thread defers `initialize_client_sessions()` until the loop is ready.
   - When the loop is later attached, `set_event_loop()` schedules a catch-up initialization so engines discovered early still get sessions initialized.
3. **Thread-safety hardening**
   - Ensure the new coordination fields are initialized before the watcher thread is started, avoiding “partially constructed object” races.

## Reproduction (original bug)

The easiest way to reproduce the original issue is to start with **one** router replica (so the stack comes up cleanly), and then scale the router deployment to **two or more** replicas. Restarting the router after an engine is already running also works most of the time.

### 0. Create a local Kind cluster (easiest local repro)

If you do not already have a Kubernetes cluster, the quickest way to reproduce locally is to create a GPU-enabled Kind cluster using the repo scripts described in:

- `tutorials/00-c-install-kind-kubernetes-env.md`

From the repo root:

```bash
# Install required tools (if you do not already have them)
bash utils/install-kind.sh
bash utils/install-kubectl.sh
bash utils/install-helm.sh

# Create a Kind cluster named "single-node-cluster" and install the NVIDIA GPU operator
bash utils/install-kind-cluster.sh
```

### 1. Install the minimal stack

Start from the basic tutorial:

- `tutorials/01-minimal-helm-installation.md`

Install the stack with the default single router replica (as in the tutorial), then scale up:

```bash
helm repo add vllm https://vllm-project.github.io/production-stack
helm upgrade --install vllm vllm/vllm-stack -f tutorials/assets/values-01-minimal-example.yaml

# After the stack is running, scale the router to 2 (or any number >= 2)
kubectl scale deployment/vllm-deployment-router --replicas=2
```

On the affected code, one of the router pods typically logs:

```text
Error initializing client sessions: 'State' object has no attribute 'event_loop'
```

## Testing

- GitHub Actions (single GPU): all tests I was able to run in my own GitHub repo passed.
- Multi-GPU tests: not run (requires 2+ GPUs).
- Upstream CI: should provide additional verification for multi-GPU scenarios and the broader matrix once execution is approved.

Manual validation was also performed:

- Local Kind cluster
- Two router replicas
- Observed clean startup and no recurrence of the `event_loop` initialization error

## Notes

- The change is intentionally small and scoped to startup coordination.
- `set_event_loop()` is only used when present (checked via `hasattr`), so other service discovery types remain unaffected.
- The fix avoids changing routing behavior: it only prevents scheduling async work before the loop exists.
- I'm aware of the effort to rewrite the router in Rust. Yet, since this is still the official implementation, I believe fixing this is a good idea in the meantime.

### DIFF
--- a/src/vllm_router/app.py
+++ b/src/vllm_router/app.py
@@ -88,11 +88,16 @@ async def lifespan(app: FastAPI):
     if hasattr(app.state, "batch_processor"):
         await app.state.batch_processor.initialize()
 
+    # Attach the running event loop early so background threads can schedule
+    # coroutines without races.
+    loop = asyncio.get_event_loop()
+    app.state.event_loop = loop
+
     service_discovery = get_service_discovery()
+    if hasattr(service_discovery, "set_event_loop"):
+        service_discovery.set_event_loop(loop)
     if hasattr(service_discovery, "initialize_client_sessions"):
         await service_discovery.initialize_client_sessions()
-
-    app.state.event_loop = asyncio.get_event_loop()
 
     yield
     await app.state.aiohttp_client_wrapper.stop()
--- a/src/vllm_router/service_discovery.py
+++ b/src/vllm_router/service_discovery.py
@@ -425,12 +425,33 @@ class K8sPodIPServiceDiscovery(ServiceDiscovery):
         self.k8s_api = client.CoreV1Api()
         self.k8s_watcher = watch.Watch()
 
+        self.prefill_model_labels = prefill_model_labels
+        self.decode_model_labels = decode_model_labels
+        # Event loop coordination: set when the FastAPI loop is available.
+        self.event_loop = None
+        self.event_loop_ready = threading.Event()
+
         # Start watching engines
         self.running = True
         self.watcher_thread = threading.Thread(target=self._watch_engines, daemon=True)
         self.watcher_thread.start()
-        self.prefill_model_labels = prefill_model_labels
-        self.decode_model_labels = decode_model_labels
+
+    def set_event_loop(self, loop: asyncio.AbstractEventLoop) -> None:
+        """
+        Attach the running FastAPI event loop so background threads can
+        safely schedule coroutines.
+        """
+        self.event_loop = loop
+        self.event_loop_ready.set()
+
+        # Ensure client sessions are initialized for any engines discovered
+        # before the loop became available.
+        try:
+            asyncio.run_coroutine_threadsafe(
+                self.initialize_client_sessions(), self.event_loop
+            )
+        except Exception as e:
+            logger.warning(f"Deferred client session init failed: {e}")
 
     @staticmethod
     def _check_pod_ready(container_statuses):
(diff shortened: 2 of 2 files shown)


ruizhang0101:
Hi Can you update the branch?

bcdonadio:
Of course. The other PR you just merged is also exactly the other one I kept in my fork as well. Just a moment.

bcdonadio:
@zerofishnoodles There. Properly cleaned and rebased.

PR #770, opened 2025-12-16
[Build][Router] Update vllm to v0.13.0

shernshiou:
The latest lmcache/lmstack-router docker image (lmcache/lmstack-router:0.1.9.dev4-ge1447afd3.d20251215) contains several CVEs including CVEs belonging to vllm v0.11.0 and also older version of starlette. Updating vllm to v0.12.0 reduces the CVEs and also update the starlette to the version without CVE. 

---

- [x] Try to classify PRs for easy understanding of the type of changes, such as `[Bugfix]`, `[Feat]`, and `[CI]`.:

### DIFF
--- a/src/vllm_router/app.py
+++ b/src/vllm_router/app.py
@@ -253,6 +253,8 @@ def initialize_all(app: FastAPI, args):
         prefill_model_labels=args.prefill_model_labels,
         decode_model_labels=args.decode_model_labels,
         kv_aware_threshold=args.kv_aware_threshold,
+        lmcache_health_check_interval=args.lmcache_health_check_interval,
+        lmcache_worker_timeout=args.lmcache_worker_timeout,
     )
 
     # Initialize feature gates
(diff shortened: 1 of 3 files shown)


ruizhang0101:
Hi Can you revert the change to the uv lock since there are too many change line in that.

shernshiou:
> Hi Can you revert the change to the uv lock since there are too many change line in that.

Hi @zerofishnoodles 
Yes, I can but can you help me understand why? Shouldn't uv.lock be consistent with pyproject.toml?

shernshiou:
@zerofishnoodles I have reset the uv.lock

bcdonadio:
I believe a production-stack release should follow this update as well, right?

ruizhang0101:
> > Hi Can you revert the change to the uv lock since there are too many change line in that.
> 
> Hi @zerofishnoodles Yes, I can but can you help me understand why? Shouldn't uv.lock be consistent with pyproject.toml?

Yes it should be. It is just that we don't want too many changes in one PR. Also since pyproject itself should be enough for the venv.

ruizhang0101:
> I believe a production-stack release should follow this update as well, right?

Yes it should be

ruizhang0101:
Thanks for the contribution again! Would you be able to fix the CI issue?

PR #784, opened 2026-01-19
Fix incorrect import path in batch processor initialization

### LINKED ISSUE #766: bug: initialize_batch_processor has an invalid import
### Describe the bug

When the flag `--enable_batch_api` is added to your startup args, vLLM will attempt to create a local storage service for persisting files used in batch requests. This argument is handled on line 186 of `src/vllm-router/app.py`. 

On line 191, initialize_batch_processor is called, which has been imported as:
`from vllm_router.services.batch_service import initialize_batch_processor`

Within `src/vllm_router/services/batch_service/__init__.py`, the following code attempts an import:

```
    if batch_processor_name == "local":
        from vllm_router.batch.local_processor import LocalBatchProcessor

        return LocalBatchProcessor(storage_path, storage)
    else:
        raise ValueError(f"Unknown batch processor: {batch_processor_name}")
```

The import of `from vllm_router.batch.local_processor import LocalBatchProcessor` is incorrect as this module does not exist. 

### To Reproduce

Add the flag `--enable_batch_api` to your startup args. You will then see an error about this import not working. 

### Expected behavior

You are able to run vLLM with local storage to enable batch mode. 

### Additional context

I believe the correct import on that line should be:
`from vllm_router.services.batch_service.local_processor import LocalBatchProcessor`

keyuchen21:
## Summary

Fixes #766

This PR corrects an incorrect import path that prevented the batch API feature from working.

## Problem

When users tried to enable the batch API with the `--enable_batch_api` flag, the router would crash with a `ModuleNotFoundError` because the import statement was referencing a non-existent module path.

**Incorrect import (line 9):**
```python
from vllm_router.batch.local_processor import LocalBatchProcessor
```

The directory `vllm_router/batch/` does not exist in the codebase.

## Solution

Changed the import to use the correct path where `LocalBatchProcessor` actually exists:

```python
from vllm_router.services.batch_service.local_processor import LocalBatchProcessor
```

## Impact

- Unblocks the batch API functionality which was completely unusable due to this import error
- Users can now successfully use the `--enable_batch_api` flag
- No breaking changes or API modifications
- Single-line fix with zero risk

## Testing

- Verified that `src/vllm_router/services/batch_service/local_processor.py` exists and contains the `LocalBatchProcessor` class
- Searched the entire codebase to confirm no other files reference the incorrect import path
- Confirmed the import statement follows Python module structure conventions

### DIFF
--- a/src/vllm_router/services/batch_service/__init__.py
+++ b/src/vllm_router/services/batch_service/__init__.py
@@ -6,7 +6,9 @@ def initialize_batch_processor(
     batch_processor_name: str, storage_path: str, storage: Storage
 ) -> BatchProcessor:
     if batch_processor_name == "local":
-        from vllm_router.services.batch_service.local_processor import LocalBatchProcessor
+        from vllm_router.services.batch_service.local_processor import (
+            LocalBatchProcessor,
+        )
 
         return LocalBatchProcessor(storage_path, storage)
     else:


keyuchen21:
@zerofishnoodles i think now it should be good to go now

                                                                                                                                                                                                                                                      
previous code:                                                                                                                                                                                                                                                   
  from vllm_router.services.batch_service.local_processor import LocalBatchProcessor                                                                                                                                                                              
                                                                                                                                                                                                                                                                  
 change to:                                                                                                                                                                                                                                                    
  from vllm_router.services.batch_service.local_processor import (                                                                                                                                                                                                
      LocalBatchProcessor,                                                                                                                                                                                                                                        
  )

PR #791, opened 2026-01-21
[Fix] Add null check for session_key in KvawareRouter

### LINKED ISSUE #735: bug: session_key args has no null check
### Describe the bug

The session_key args under KvawareRouter has no null check before accessing it.

### To Reproduce

Run the vllm production stack with below args
--host                                                                                                                                                                               │
│       0.0.0.0                                                                                                                                                                              │
│       --port                                                                                                                                                                               │
│       8000                                                                                                                                                                                 │
│       --service-discovery                                                                                                                                                                  │
│       k8s                                                                                                                                                                                  │
│       --k8s-namespace                                                                                                                                                                      │
│       vllm                                                                                                                                                                                 │
│       --k8s-label-selector                                                                                                                                                                 │
│       environment=test,release=test                                                                                                                                                        │
│       --routing-logic                                                                                                                                                                      │
│       kvaware                                                                                                                                                                              │
│       --engine-stats-interval                                                                                                                                                              │
│       15                                                                                                                                                                                   │
│       --request-stats-window                                                                                                                                                               │
│       60                                                                                                                                                                                   │
│       --lmcache-controller-port                                                                                                                                                            │
│       9000

### Expected behavior

_No response_

### Additional context

_No response_

keyuchen21:
## Summary

Fixes #735

This PR adds a null check for the `session_key` argument in `KvawareRouter.__init__()`.

## Problem

When users run kvaware routing without specifying `--session-key`, the router fails with an unclear error because `session_key` is accessed without validation.

## Solution

Added a null check that raises a clear `ValueError` if `session_key` is None, matching the existing behavior in `SessionRouter`:

```python
if session_key is None:
    raise ValueError("KvawareRouter must be initialized with a session_key")
```

## Testing

- Verified syntax is correct
- The fix follows the same pattern used in `SessionRouter` (line 191-192)

### DIFF
--- a/docs/source/use_cases/semantic-router-integration.rst
+++ b/docs/source/use_cases/semantic-router-integration.rst
@@ -74,21 +74,29 @@ Identify the ClusterIP and port of your router Service:
 Step 2: Deploy vLLM Semantic Router
 ------------------------------------
 
-Follow the official `Install in Kubernetes <https://vllm-semantic-router.com/docs/installation/kubernetes>`_ guide with the updated configuration.
+Follow the official `Install in Kubernetes <https://vllm-semantic-router.com/docs/installation/k8s/ai-gateway>`_ guide with the updated configuration.
+
+Deploy vLLM Semantic Router using Helm:
 
 .. code-block:: bash
 
-   # Deploy vLLM Semantic Router manifests
-   kubectl apply -k deploy/kubernetes/ai-gateway/semantic-router
+   # Deploy vLLM Semantic Router with custom values from GHCR OCI registry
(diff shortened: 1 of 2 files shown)


ruizhang0101:
Hi, Can you confirm this error still exists?

keyuchen21:
@zerofishnoodles 
- I tried to reproduce with #735  here is the result, please see screenshot
- it looks good to me

image

ruizhang0101:
Thanks for the screenshot. I meant will there be an error without this PR. Because kvaware router doesn't have to have a session key.

keyuchen21:
@zerofishnoodles  you are right........
Your PR #741 Fixed It ....... I did not realized it..

Issue #807, opened 2026-01-30
bug: serviceEngine.probe.exec field does not remove httpGet


emanuelecassese:
### Describe the bug

hi,
i have to add a liveness probe to the inference pod to check the model responses
in the _helpers.tpl file, we are using toYaml function

<img width="494" height="422" alt="Image" src="https://github.com/user-attachments/assets/6c5e0ddf-adcd-4c9a-b28c-bbd8a7c7324d" />

And in the values.yaml file there are values for the httpGet field in the probes

<img width="276" height="886" alt="Image" src="https://github.com/user-attachments/assets/c9683409-5aa9-427d-b40b-57d90f0cdd1a" />

Like this, if you add in the custom values file the serviceEngineSpec.livenessProbe.**EXEC** field to use a custom command for the probe, the helm chart simply add the field and the final manifest cannot be applied due to overlapping of the 2 fields httpGet and exec

<img width="596" height="364" alt="Image" src="https://github.com/user-attachments/assets/f761df79-473d-4c84-96ae-8a6e642cc219" />

### To Reproduce

Add the exec field in one of the 3 probe in serviceEngineSpec

### Expected behavior

The Helm Chart must exclude one of the field because the probe cannot manage 2 handler type

PR #808, opened 2026-01-30
[BugFix][Feat]: fix serviceEngineSpec probe field and improve probe management in helm template

emanuelecassese:
FIX https://github.com/vllm-project/production-stack/issues/807

### Values File Modification: 
The serviceEngineSpec.probe.httpGet section in the values file has been commented 
Added equivalent values to the template defaults to prevent the httpGet field from being automatically included in the deployment template.
        
### Helper File Update:
Enhanced the define.probes in the helpers file.
Attempted to accommodate all possible fields to closely align with the Kubernetes probe definitions.

## Impact
These changes ensure that the httpGet field does not get unnecessarily added, thus making the Helm chart more compliant with expected Kubernetes configurations.

### DIFF
--- a/helm/templates/_helpers.tpl
+++ b/helm/templates/_helpers.tpl
@@ -84,24 +84,154 @@ Define additional router ports
 Define startup, liveness and readiness probes
 */}}
 {{- define "chart.probes" -}}
-{{-   if .Values.servingEngineSpec.startupProbe  }}
+{{- if .Values.servingEngineSpec.startupProbe }}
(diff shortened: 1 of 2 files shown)


emanuelecassese:
These are some of the test values file i've use to test my changes

[tests.yml](https://github.com/user-attachments/files/24967195/tests.yml)

emanuelecassese:
Close to fix the grpc and tcpsocket problems

PR #809, opened 2026-01-30
[BugFix][Feat]: fix serviceEngineSpec probe field and improve probe management in helm template

emanuelecassese:
FIX https://github.com/vllm-project/production-stack/issues/807
### Values File Modification:

The serviceEngineSpec.probe.httpGet section in the values file has been commented
Added equivalent values to the template defaults to prevent the httpGet field from being automatically included in the deployment template.
### Helper File Update:

Enhanced the define.probes in the helpers file.
Attempted to accommodate all possible fields to closely align with the Kubernetes probe definitions.

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -266,9 +266,9 @@ servingEngineSpec:
       60
       # -- Configuration of the Kubelet http request on the server
     httpGet:
-    # -- Path to access on the HTTP server
+      # -- Path to access on the HTTP server
       path: /health
-    # -- Name or number of the port to access on the container, on which the server is listening
+      # -- Name or number of the port to access on the container, on which the server is listening
       port: 8000
 
   # -- Liveness probe configuration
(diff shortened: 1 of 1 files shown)


emanuelecassese:
These are some of the test values i've use to test my changes
[tests.yml](https://github.com/user-attachments/files/24969049/tests.yml)

helm/templates/_helpers.tpl
```diff
@@ -84,24 +84,148 @@ Define additional router ports
...
+        value: {{ .value }}
+    {{- end }}
+    {{- end }}
+    {{- if .httpGet.host }}
+    host: {{ .httpGet.host }}
+    {{- end }}
+    {{- if .httpGet.scheme }}
+    scheme: {{ .httpGet.scheme }}
+    {{- end }}
+  {{- end }}
```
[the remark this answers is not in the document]
emanuelecassese: Done!

ruizhang0101 (changes requested):
Hi! Thanks for bringing this up. 
For each specific type, could you use toyaml since there are way too many duplication?

emanuelecassese:
Hi @ruizhang0101, 
thanks for the idea
I have improved the chart adding a probe template in the helpers to avoid duplication
I would like to keep the explicit field definition instead of using toYaml

helm/values.yaml
```diff
@@ -265,9 +265,9 @@ servingEngineSpec:
       60
       # -- Configuration of the Kubelet http request on the server
     httpGet:
-      # -- Path to access on the HTTP server
+    #  # -- Path to access on the HTTP server
```
ruizhang0101: Can you remove these unnecessary modifications? Same with line R270, R283, R285

emanuelecassese: Oh yeah, sorry about that
Done

PR #810, opened 2026-01-30
feat(helm): add standard Kubernetes labels to deployments and services

### LINKED ISSUE #671: feature: Add model name to appication instance labels
### Describe the feature

The goal is to correctly label the vllm engine to be recognized as an application instance

_No response_

### Additional context

_No response_

keyuchen21:
## Summary
- Add `app.kubernetes.io/*` labels to enable proper recognition by Kubernetes tools like Lens
- Labels added to engine deployments, services, and scaled objects
- Labels added to router deployment and service

## Changes
- Added `chart.engineStandardLabels` and `chart.routerStandardLabels` helpers in `_helpers.tpl`
- Applied standard labels to all relevant resources in:
  - `deployment-vllm-multi.yaml`
  - `deployment-router.yaml`
  - `service-vllm.yaml`
  - `service-router.yaml`
  - `scaledobject-vllm.yaml`

## Test plan
- [x] `helm template` renders all resources correctly
- [x] `helm lint` passes with no errors
- [x] Existing custom labels preserved (`model`, `helm-release-name`)

Closes #671

### DIFF
--- a/helm/templates/deployment-router.yaml
+++ b/helm/templates/deployment-router.yaml
@@ -5,8 +5,8 @@ metadata:
   name: "{{ .Release.Name }}-deployment-router"
   namespace: {{ .Release.Namespace }}
   labels:
-  {{- include "chart.routerLabels" . | nindent 4 }}
   {{- include "chart.routerStandardLabels" (dict "releaseName" .Release.Name "chartName" .Chart.Name) | nindent 4 }}
+  {{- include "chart.routerLabels" . | nindent 4 }}
 spec:
 {{- if .Values.routerSpec.autoscaling.enabled }}
   replicas: {{ .Values.routerSpec.autoscaling.minReplicas }}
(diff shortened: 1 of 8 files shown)


helm/templates/deployment-router.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
   {{- include "chart.routerLabels" . | nindent 4 }}
+  {{- include "chart.routerStandardLabels" (dict "releaseName" .Release.Name "chartName" .Chart.Name) | nindent 4 }}
```
ruizhang0101, line 8: Hi, Can you put the standard labels ahead of custom labels so that the custom labels can override the standard label if they have the same keys? same with other places defining standard labels

ruizhang0101 (changes requested):
Hi Can you

helm/templates/deployment-router.yaml
keyuchen21: Good point!!!!, just updated it.

ruizhang0101 (approved):
LGTM :)) Thanks for the contribution!

PR #832, opened 2026-02-11
fix: make --log-level CLI argument actually control router log levels

### LINKED ISSUE #817: bug: Router log level is unchangeable, and default in log.py is DEBUG
### Describe the bug

Hello,

We're using the vllm-production stack with a router in production, with thousands of requests each minute and our Kubernetes logging stack got saturated with logs over 150GB per hour!

After investigating, we realized the router is the main culprit as it was logging every request in DEBUG and we tried for hours to drop it to INFO or WARNING and nothing worked. It seems the value is hardcoded.

This line in app.py seems to be not cheageable from DEBUG https://github.com/vllm-project/production-stack/blob/17f789ef55bc2373ada7cab657bd267a8e5659a4/src/vllm_router/log.py#L44
We had to patch it via a configmap.

We then realized even log level INFO was too verbose and still saturating our logging systems. It's due to uvicorn's log level that is also not settable via the CLI. 

We had to patch also this line https://github.com/vllm-project/production-stack/blob/17f789ef55bc2373ada7cab657bd267a8e5659a4/src/vllm_router/app.py#L350
and patch it to uvicorn.run(app, host=args.host, port=args.port, log_level=args.log_level)

Is there a way to fix this?

### To Reproduce

Run the router (docker image lmcache/vllm-openai) and then:
- Make a call to it to see DEBUG logs appear. No variable can be set to drop it to INFO.
- Try setting log level to ERROR or WARNING and make a call. Uvicorn still logs every request with INFO.

### Expected behavior

_No response_

### Additional context

_No response_

keyuchen21:
## Summary

- The `--log-level` CLI argument existed but was completely unused — all application loggers hardcoded `DEBUG` level via `init_logger()`, and `uvicorn.run()` was called without a `log_level` parameter. This caused massive log volumes in production (reported: 150GB/hour in #817).
- Added `set_log_level()` function in `log.py` that retroactively updates all loggers created at import time (before args are parsed), changed the default from `DEBUG` to `INFO`, and wired the CLI arg into both application loggers and uvicorn.

## Changes

- **`src/vllm_router/log.py`**: Add `set_log_level()` to update all tracked loggers, default to `INFO` instead of `DEBUG`, track loggers in a module-level list
- **`src/vllm_router/app.py`**: Call `set_log_level(args.log_level)` after `parse_args()`, pass `log_level` to `uvicorn.run()`
- **`src/vllm_router/parsers/parser.py`**: Update `--log-level` help text to reflect it now controls both the router and uvicorn

Fixes #817

## Test plan

- [x] Existing unit tests pass (32/32)
- [x] `--log-level warning` suppresses DEBUG/INFO: all 13 application loggers set to WARNING, DEBUG/INFO messages silenced, WARNING messages still visible
- [x] Uvicorn access logs respect the level: `uvicorn.Config` receives `log_level="warning"`, both `uvicorn.error` and `uvicorn.access` loggers set to WARNING

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ dependencies = [
     "kubernetes==32.0.0",
     "numpy==1.26.4",
     "prometheus-client==0.21.1",
-    "python-multipart==0.0.20",
+    "python-multipart==0.0.22",
     "sentry-sdk[fastapi]==2.27.0",
     "uhashring==2.3",
     "uvicorn==0.34.0",
--- a/src/vllm_router/requirements.txt
+++ b/src/vllm_router/requirements.txt
@@ -1,5 +1,5 @@
 aiofiles==24.1.0
-aiohttp[speedups]==3.13.0
+aiohttp[speedups]==3.13.3
 fastapi==0.128.0
 kubernetes==32.0.0
 numpy==1.26.4
@@ -10,7 +10,7 @@ opentelemetry-exporter-otlp==1.28.0
 opentelemetry-sdk==1.28.0
 prometheus_client==0.21.1
 psutil==7.0.0
-python-multipart==0.0.20
+python-multipart==0.0.22
 sentry-sdk[fastapi]==2.27.0
 uhashring==2.3
 uvicorn==0.34.0


keyuchen21:
@ruizhang0101  happy to help!

I opened #835 for it. please take a look if it is looks ok
https://github.com/vllm-project/production-stack/pull/835

PR #836, opened 2026-02-17
fix: filter hop-by-hop headers from streaming responses

### LINKED ISSUE #827: bug: Istio compression messes up the router-engine connection and user receives decompressed body with content-encoding: "gzip"
### Describe the bug

Hello. Since we enabled istio compression in our Kubernetes cluster, all vLLM calls started failing because the response returned by the router is decompressed with a content-encoding: "gzip".

I tried to investigate as much as I can, and this might not be the reason but:
https://github.com/vllm-project/production-stack/blob/main/src/vllm_router/services/request_service/request.py#L776
This line sends the same content-encoding "gzip" it receives from the engine's istio sidecar back to the caller. However this line https://github.com/vllm-project/production-stack/blob/8f2ca50ce7b01f0a3fdca967792cbbd0fceee49a/src/vllm_router/services/request_service/request.py#L772 calls a aiohttp session where by default auto_decompress is enabled! As you can see here
https://docs.aiohttp.org/en/stable/client_reference.html

So that explains why the users receives a decompressed response with a content "gzip". I believe the user<>router istio doesn't do anything since it believes that it's already compressed.

I don't know what a proper fix should be. I will override this line https://github.com/vllm-project/production-stack/blob/8f2ca50ce7b01f0a3fdca967792cbbd0fceee49a/src/vllm_router/aiohttp_client.py#L31 with a auto_decompress=False and see if that solves the issue.

### To Reproduce

Have istio compression enabled in your k8s cluster and use a router to forward your requests to the engines. You will receive a decompressed body with content-type "gzip".

### Expected behavior

_No response_

### Additional context

_No response_

keyuchen21:
## Summary

- Filter out `content-length`, `content-encoding`, `transfer-encoding`, and `connection` headers from streaming response headers in `route_general_request()`
- Fixes #827: when Istio compression is enabled, aiohttp's `auto_decompress=True` decompresses response bodies but the original `content-encoding: gzip` header was still forwarded, causing clients to receive decompressed data labeled as gzip
- Matches the existing pattern already used in `route_general_transcriptions` (PR #733)

## Test plan

- [x] Existing tests pass (`pytest src/tests/`)
- [ ] Manual verification: with Istio compression enabled, streaming responses no longer carry a stale `content-encoding: gzip` header

### DIFF
--- a/.github/workflows/helm-release.yml
+++ b/.github/workflows/helm-release.yml
@@ -11,12 +11,13 @@ on:
       release_type:
         description: 'What to release'
         required: true
-        default: 'both'
+        default: 'all'
         type: choice
         options:
-          - docker_only
+          - router_only
+          - crd_only
           - helm_only
-          - both
+          - all
 
 jobs:
   release:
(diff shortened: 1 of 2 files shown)


src/vllm_router/services/request_service/request.py
```diff
@@ -84,6 +84,13 @@
     "trailer",
 }
 
+_HEADERS_TO_STRIP_FROM_STREAMING_RESPONSE = {
```
ruizhang0101: Can you change this to a general name since the transcription one is not streaming?

ruizhang0101:
Hi Could you sign off the commit?

src/vllm_router/services/request_service/request.py
keyuchen21: Done, renamed to `_HEADERS_TO_STRIP_FROM_RESPONSE`.

keyuchen21:
@ruizhang0101  done....

PR #844, opened 2026-02-20
feat: add --root-path CLI option for hosting router under a subpath

### LINKED ISSUE #760: feature: support configurable FastAPI root path
### Describe the feature

I want to be able to host vLLM under a subpath (e.g. example.com/vllm) instead of dedicating a whole subdomain.

`vllm serve` supports configuring the FastAPI root path. The router should expose the same option so it can run under a subpath as well.

### Additional context

https://docs.vllm.ai/en/stable/cli/serve/#-root-path

keyuchen21:
## Summary
- Adds `--root-path` CLI argument to the router, allowing it to be hosted under a subpath (e.g. `example.com/vllm`) instead of requiring a dedicated subdomain
- Passes `root_path` through to `uvicorn.run()`, which sets it on the ASGI app automatically
- Adds Helm chart support via `routerSpec.rootPath`

Closes #760

## Changes
- `src/vllm_router/parsers/parser.py` — new `--root-path` argument (default: `""`)
- `src/vllm_router/app.py` — pass `root_path` to `uvicorn.run()`
- `helm/values.yaml` — add `rootPath: ""` under `routerSpec`
- `helm/templates/deployment-router.yaml` — conditionally pass `--root-path` arg

## Test plan
- [x] Verified `--root-path /vllm` parses correctly
- [x] Verified default empty string preserves existing behavior
- [x] Existing tests pass (28/29, one pre-existing failure unrelated to this change)

### DIFF
--- a/.github/workflows/router-e2e-test.yml
+++ b/.github/workflows/router-e2e-test.yml
@@ -137,8 +137,8 @@ jobs:
             --chunk-size 128 \
             --verbose \
             --result-dir /tmp/k8s-discovery-routing-results-pr-${{ github.event.pull_request.number || 'main' }} \
-            --timeout 10
-        timeout-minutes: 10
+            --timeout 20
+        timeout-minutes: 20
 
       - name: Archive k8s discovery routing test results
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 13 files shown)


keyuchen21:
#760 is asking for it.
@ruizhang0101  I think it should be useful when in production environments where services are deployed behind a reverse proxy, API gateway, or Kubernetes Ingress that routes multiple services under a shared domain using path-based routing.

@kkmarv if you want to elaborate on it more

kkmarv:
@keyuchen21 yes, exactly what you said.

In (Kubernetes) environments where one cannot easily obtain new domains, path-based routing is the only alternative if you want to expose multiple services.

The same goes for dev environments too. Sometimes developers cannot create new domains within a company network due to compliance reasons.

I'd certainly enjoy it if vllm could offer the extra flexibility, so thank you for putting the effort into this!

PR #850, opened 2026-02-23
[Router]: image edit routes multi-part form request

### LINKED ISSUE #790: feature: vllm omni
### Describe the feature

Integrate vllm omni to use qwen image through kubernetes

_No response_

### Additional context

_No response_

nmiguel:
- [X] Make sure the code changes pass the [pre-commit](https://github.com/vllm-project/production-stack/blob/main/CONTRIBUTING.md) checks.
- [X] Sign-off your commit by using <code>-s</code> when doing <code>git commit</code>
- [X] Try to classify PRs for easy understanding of the type of changes, such as `[Bugfix]`, `[Feat]`, and `[CI]`.

----

The current version of the Router technically supports both the image generation and image edit API routes, but a correct request to the image edit route will fail, as it's the only request in the Openai API that uses a multi-part form request instead of application/json. 

This PR attempts creates a new function for routing an image edit request. I tried to mimic the code of "route_general_request" as much as possible, but it still felt different enough and specific enough that I opted for a new function, rather than complicating the logic of route_general_request to support both multi-part form and json. 

Please let me know if there are any changes I should make.

This should close #790

### DIFF
--- a/src/vllm_router/services/request_service/request.py
+++ b/src/vllm_router/services/request_service/request.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# --- Request Processing & Routing ---
 import json
 import os
 import time
@@ -20,6 +19,9 @@ import uuid
 from typing import Optional
 
 import aiohttp
+
+# --- Request Processing & Routing ---
+from aiohttp import FormData
 from fastapi import BackgroundTasks, HTTPException, Request, UploadFile
 from fastapi.responses import JSONResponse, StreamingResponse
 from requests import JSONDecodeError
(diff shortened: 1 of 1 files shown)


ruizhang0101 (commented):
Hi, thanks for the contribution! It looks good to me. Just wondering, the route_general_transcriptions is also multi-part form request. Is it possible to merge them to a function for general form-style request?

nmiguel:
@ruizhang0101 PTAL, the form parsing is mostly specific to each request, and I wanted to maintain the new endpoint's passing through of the body as I don't see a point in processing it at all, this is the result. Let me know if you have any other changes you want. 
Also, of note: the body of the shared function is exactly what was already present in the transcription function body.

ruizhang0101 (approved):
Yeah it looks good to me! Could you merge the latest main, and then I think it is good to go :)

nmiguel:
@ruizhang0101 done. I think the CI/CD has been randomly failing.

Issue #640, opened 2025-08-11
[Roadmap] vLLM Production Stack roadmap for 2025 Q3


YuhanLiu11:
## CRD and Gateway Inference Extension

- [ ] (P0) Scale down to zero with KEDA and CRD integration (@Romero027 ) 
- [ ] (P0) Implement disaggregated prefill in CRD 
- [ ] (P0) Implementing KV cache aware and prefix aware routing logic in gateway inference extension 
- [ ] (P0) Implement disaggregated prefill in Gateway Inference Extension

## Router “frontend“

- [ ] (P2) Router performance enhancements
     - [ ] Nuitka compilation for the current router
     - [ ] Rust/Go/Nginx-based frontend for router

## Router core logic

- [ ] (P0) XpYd support 
- [ ] (P2) Load balancing for disaggregated prefill
- [ ] (P2) Priority routing
- [ ] (P2) Routing to external providers like OpenAI or Anthropic
- [ ] (P2) Request migration when the vLLM instance fails

## CI/CD and misc.

- [ ] (P0) Release bot to automatically release new versions (helm chart + k8s controller packages + docker images)
- [ ] (P2) Github actions for building router docker images for different architectures
- [ ] (P2) Support for multi-modality models

nejch:
@YuhanLiu11 I just opened 2 PRs (https://github.com/vllm-project/production-stack/pull/776, https://github.com/vllm-project/production-stack/pull/775) that both are somewhat related to items on this roadmap.

However I just noticed https://github.com/vllm-project/router based on sglang. Before I continue with follow-up PRs, do you have any info on how the two (`vllm_router`) in here and `vllm-router` in the other are related to each other? Do you plan to drop the router in here and switch to the new one?

PR #852, opened 2026-02-26
[Feat][Router] Add Priority routing

ikaadil:
Adds a new **priority** routing option that routes requests to the backend with the **lowest QPS** (least loaded).

Addresses the Q3 roadmap https://github.com/vllm-project/production-stack/issues/640

## Changes

- **PriorityRouter**: New router that uses existing `_qps_routing()` to select the least-loaded endpoint.
- **CLI / config**: `--routing-logic priority` and `priority` in dynamic config and Helm.
- **Docs**: Router README, Helm README, and semantic-router tutorial updated.
- **Tests**: Unit tests for PriorityRouter (lowest QPS, no stats, equal QPS).
- **Edge case**: PriorityRouter raises `ValueError` when given no endpoints.

## Usage

`--routing-logic priority`

---

<details>

<ul>
</ul>

<h3>Code Quality</h3>

<ul>
</ul>

, Shaoting-Feng or ApostaC.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 - 📄 [Official documentation](https://docs.vllm.ai/projects/production-stack) released for production-stack!
 - ✨ [Cloud Deployment Tutorials](https://github.com/vllm-project/production-stack/blob/main/tutorials) for Lambda Labs, AWS EKS, Google GCP are out!
-- 🛤️ 2025 Q1 roadmap is released! [Join the discussion now](https://github.com/vllm-project/production-stack/issues/26)!
+- 🛤️ 2026 roadmap is released! [Join the discussion now](https://github.com/vllm-project/production-stack/issues/855)!
 - 🔥 vLLM Production Stack is released! Check out our [release blogs](https://blog.lmcache.ai/2025-01-21-stack-release) posted on January 22, 2025.
 
 ## Community Events
--- a/helm/Chart.yaml
+++ b/helm/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.1.9
+version: 0.1.10
 
 maintainers:
   - name: apostac


ruizhang0101 (commented):
Hi, thanks for the PR. I think the qps routing is semantically different from the priority routing. In terms of the priority routing, could you elaborate the use case of this routing given that vllm itself has priority mechanism.

ikaadil:
> Hi, thanks for the PR. I think the qps routing is semantically different from the priority routing. In terms of the priority routing, could you elaborate the use case of this routing given that vllm itself has priority mechanism.

Thanks for the review. Here’s a short clarification on semantics and use case.

**1. QPS vs “priority” (semantics)**  
I agree that what this implements is **QPS-based / least-loaded** routing (choose the backend with lowest QPS), not “priority” in the usual sense of (a) request priority (high vs low) or (b) endpoint tier (prefer A then overflow to B). So the current name `priority` is a bit overloaded. I’m happy to rename this option to something like **`least_loaded`** or **`load_based`** so it’s clear we mean “route to the least-loaded backend” and avoid confusion with vLLM’s request-level priority. If you have a preferred name, I can align with that.

**2. Use case vs vLLM’s priority**  
- **vLLM’s priority** lives **inside** a single backend: it decides **which request** runs next (FCFS, or priority scheduling when enabled), preemption, batching. It does **not** choose which backend gets the request.  
- **This option** lives at the **router**: when there are **multiple** vLLM backends (e.g. multiple pods), it chooses **which backend** gets the request. Right now it does that by picking the backend with **lowest QPS** (least loaded), so we’re doing **load-aware placement** across instances.

So they’re complementary: the router decides **which instance** gets the request (this PR: least-loaded by QPS); vLLM then decides **order and priority** of requests on that instance. The use case is multi-backend deployments where you want to avoid sending more traffic to already-busy instances (e.g. reduce queue depth and latency) instead of blindly round-robinning.

I can add this distinction and use case to the PR description and docs, and rename the option to `least_loaded` (or your preferred name) if that works for you.

codesun8:
Hello, I think there is redundancy in the logic between this code and the session handling. During the session invocation process, even if no session ID is specified, it still goes through the _qps_routing method and routes based on QPS.

codesun8:
I believe there is a lack of a routing strategy based on request count. If an abnormal request occurs, it can easily cause the pod to become blocked.

nejch:
I think there's value in having something like this (essentially just an explicit QPS routing without session keys), simply because this is already the behavior of session-based routing when no session headers are used. I didn't actually know session routing falls back to this. But maybe there's a smarter way to ensure pods don't get blocked, not sure.

> I believe there is a lack of a routing strategy based on request count. If an abnormal request occurs, it can easily cause the pod to become blocked.

@codesun8 this is already the case for session-based routing when going with the QPS fallback, no?

PR #870, opened 2026-03-05
[FEAT][Helm] Add configurable NodePort support for router service

### LINKED ISSUE #763: Add support for configurable NodePort router service and safe local overrides in Helm chart
### Describe the feature

When trying to customize the `routerSpec` to use a NodePort service and set a fixed `nodePort`, I ran into several issues:

1. The chart does not support defining a `nodePort` value in `values.yaml`.  
   As a result, every `helm upgrade` recreates the router service and assigns
   a new (random) NodePort—even when no configuration has changed.

2. Setting `routerSpec.serviceType` to `NodePort` forces Kubernetes to replace
   the existing service with a new one of type ClusterIP during upgrades.

I resolved these issues locally by extending the Helm chart to support a
`routerSpec.nodePorts` field and adding safe conditional checks in the service
template. This allows the router service to:

- Keep a stable NodePort across upgrades  
- Avoid being replaced or recreated unnecessarily  
- Allow users to define custom NodePort values cleanly from `values.yaml`

I would like to submit a pull request adding this improved NodePort support to
the upstream chart.

### Additional context

_No response_

ardecode:
This PR adds support for pinning stable NodePort values in the router service,
  addressing the issue where `helm upgrade` causes Kubernetes to reassign random
  ports on each update.

  Two new optional fields are added under `routerSpec.nodePorts`:
  - `router`: NodePort for the main router service port
  - `lmcache`: NodePort for the lmcache port (9000)

  Both fields default to `""`, preserving existing behavior — if unset,
  Kubernetes assigns ports randomly as before. The `nodePort` field is only
  rendered in the Service manifest when `serviceType` is `NodePort` and an
  explicit value is provided.

  Fixes #763

  ---

### DIFF
--- a/examples/disaggregated_prefill_orchestrated/README.md
+++ b/examples/disaggregated_prefill_orchestrated/README.md
@@ -1,0 +1,149 @@
+# Disaggregated Prefill Orchestrated Mode
+
+This example demonstrates the `disaggregated_prefill_orchestrated` routing mode, which orchestrates the prefill→decode flow internally within the router. Unlike the LMCache-based `disaggregated_prefill` mode, this approach uses vLLM's native `kv_transfer_params` for KV cache handoff.
+
+## Overview
+
+In this mode, the router:
+
+1. Receives a client request
+2. Sends the request to the **prefill endpoint** with `kv_transfer_params` and `max_tokens=1`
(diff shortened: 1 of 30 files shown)


ardecode:
@ruizhang0101 , PR is ready for review!

PR #873, opened 2026-03-06
fix(benchmark/multi-round-qa): fix TTFT None Type crash caused by reasoning models   (reasoning_content)

brokedba:
## Error with current version of `multi-round-qa-.py`

```
File "./multi-round-qa.py", line 153, in _async_launch_request
    ttft=first_token_time - start_time,
         ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~
TypeError: unsupported operand type(s) for -: 'NoneType' and 'float'
```

## Root Cause

The streaming loop skipped chunks where `tok.choices` was empty via an early `continue`. 
- In newer vLLM versions, the final usage chunk (carrying `completion_tokens` / `prompt_tokens`) arrives with an empty `choices` list, so token counts were never captured. 
- **Additionally**: 
Reasoning models (e.g. DeepSeek-R1, gpt-oss-20B) emit `reasoning_content` instead of `content` on the delta, so `first_token_time` was never set — causing the `ttft` calculation to crash.

## Changes brought by this PR

- Initialized `tokens_out = 0` and `tokens_prefill = 0` before the loop as safe defaults
- Moved usage extraction (`tok.usage`) before the `tok.choices` guard so it is captured from the final chunk
- Added `reasoning_content` fallback via `getattr` to support reasoning models
- Added null-safe TTFT calculation: falls back to `start_time` if no token was ever received, returning `ttft=0` instead of crashing

### DIFF
--- a/benchmarks/multi-round-qa/multi-round-qa.py
+++ b/benchmarks/multi-round-qa/multi-round-qa.py
@@ -126,6 +126,8 @@ class RequestExecutor:
         start_time = time.time()
         first_token_time = None
         words = ""
+        tokens_out = 0
+        tokens_prefill = 0
 
         response = await self.client.chat.completions.create(
             messages=messages,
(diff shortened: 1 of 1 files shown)


brokedba:
After updating the branch and pushed again the CI is running low on resources . I tried twice and each time one of the tests was failing out of time out . I think it's a resource starvation issue . Will retry tonight.

```
# failing job steps 
1st try: Secure-Minimal-Example 
2nd try : Two-Pods-Minimal-Example
```

Issue #855, opened 2026-02-27
[Roadmap] vLLM Production Stack 2026 Roadmap


ruizhang0101:
## Router

- [ ] (P0) XpYd disaggregated prefill support
- [ ] (P1) Routing to external providers like OpenAI or Anthropic
- [ ] (P1) Router-side request queuing support
- [ ] (P2) Implement predictive routing based on future workloads
- [ ] (P2) Implement priority routing
- [ ] (P2) Improve router performance
  - [ ] Compile current router with Nuitka
  - [ ] Prototype Rust/Go/Nginx-based router frontend

## Advanced Deployment

- [ ] (P0) Advanced autoscaling based on modern inference metrics
- [ ] (P0) Implement KV-cache–aware and prefix-aware routing logic in Gateway Inference extension
- [ ] (P0) Implement disaggregated prefill support in CRD
- [ ] (P0) Integrate KEDA-based scaling in CRD
- [ ] (P1) Make existing deployment methods (Helm/CRD/Gateway Inference) more Kubernetes-native
- [ ] (P2) Support automatic configuration of engine parameters

## Support for Modern Workloads

- [ ] (P0) Support deployment with vLLM Omni
- [ ] (P1) Intelligent routing for agent workloads and multimodal models
- [ ] (P2) End-to-end deployment for multimodal models
- [ ] (P2) First-class support for RL workloads

## Safety and Security
- [ ] (P0) Request migration when the vLLM instance fails
- [ ] (P1) Ship router extension config with pre-defined safety policies, rate limits, and logging enabled
- [ ] (P1) Add optional PII detection in the request path

## CI/CD and Misc

- [ ] (P0) Release bot for automated releases (Helm chart, K8s controller packages, Docker images)
- [ ] (P2) GitHub Actions for building router Docker images for multiple architectures
- [ ] (P2) Tutorials covering more cloud platforms, models, and features

If any of the items you wanted are not on the roadmap, your suggestion and contribution are strongly welcomed! Please feel free to comment in this thread, open a feature request, or create an RFC.

_Happy vLLMing!_

surak:
Forgive me the ignorance, but why there's a python vllm-router here and a completely different one in  https://github.com/vllm-project/router ? This makes it confusing to choose in a production stack.

ruizhang0101:
> Forgive me the ignorance, but why there's a python vllm-router here and a completely different one in https://github.com/vllm-project/router ? This makes it confusing to choose in a production stack.

[...]

PR #876, opened 2026-03-07
[FEAT][RFC] Add router-side request queuing proposal

ardecode:
Refs #855

## Summary

This PR adds an RFC for router-side request queuing in the vLLM router.

The proposal introduces router-side admission control in front of backend replicas to:
- smooth bursty traffic
- keep backend `vllm:num_requests_waiting` shallow
- provide bounded queueing behavior and explicit `429` overload responses

This PR is proposal-only.

## What is included

- Add `proposals/router-side-request-queuing.md`
- Define the architecture for router-side request queueing
- Scope Phase 1 to `roundrobin` routing only
- Describe follow-up phases for `session`, then `kvaware` / `prefixaware`

## Key design points in the RFC

- Per-model FIFO queueing
- Per-endpoint reservations / leases to avoid thundering herd behavior
- Admission based primarily on backend `num_requests_waiting`
- Lease release on first response chunk, or on pre-first-token error/cancel
- `429` for queue full, queue timeout, and pinned-endpoint overload
- `?id=` requests are immediate admit/reject only in Phase 1

## Why Phase 1 is limited to roundrobin

The RFC intentionally starts with `roundrobin` support only so the queue core can land without simultaneously deciding:
- session stickiness vs fallback policy
- LMCache locality vs FIFO fairness for `kvaware` / `prefixaware`

Those policy questions are deferred to later phases after the core queueing mechanism is validated.

---

- [ ] Make sure the code changes pass the [pre-commit](https://github.com/vllm-project/production-stack/blob/main/CONTRIBUTING.md) checks.
- [ ] Sign-off your commit by using <code>-s</code> when doing <code>git commit</code>
- [ ] Try to classify PRs for easy understanding of the type of changes, such as `[Bugfix]`, `[Feat]`, and `[CI]`.

### DIFF
--- a/proposals/router-side-request-queuing.md
+++ b/proposals/router-side-request-queuing.md
@@ -1,0 +1,231 @@
+# Router-Side Request Queuing for vLLM Router
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+- [Proposal](#proposal)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Implementation Timeline / Phases](#implementation-timeline--phases)
+- [References](#references)
+
+## Summary
+
+This proposal adds router-side request queuing to the vLLM router as an admission-control layer in front of backend replicas. The goal is to smooth bursty traffic, keep backend `vllm:num_requests_waiting` shallow, and provide explicit overload behavior instead of immediately forwarding every request to already saturated replicas. The initial implementation is intentionally scoped to `roundrobin` routing only, with a design that can later be extended to `session`, `kvaware`, and `prefixaware` routing.
+
+## Motivation
+
+Today, the router immediately selects a backend and proxies the request. This works when capacity is available, but it provides no router-side backpressure when replicas are already overloaded. In practice, this means the system relies entirely on the backend's internal queue and offers no bounded waiting policy, no router-level fairness, and no clear `429` behavior for clients.
+
+- **Why is this feature needed?** Bursty workloads can overwhelm backend replicas before autoscaling or normal scheduling has time to react. A router-side queue allows the stack to absorb short spikes and reject excess work cleanly when the queue budget is exhausted.
+- **What current limitations does it alleviate?** The router currently has no admission control, no bounded wait time, and no explicit queue metrics for incoming requests.
+
+### Goals
+
+- Add bounded router-side queuing for inference requests before they are dispatched to a backend replica
+- Use backend `num_requests_waiting` as the primary overload signal to decide when to queue or admit
+- Preserve FIFO behavior at the model level for the initial implementation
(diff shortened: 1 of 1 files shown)


ardecode:
@ruizhang0101 can you kindly take a look at this PR? TIA!

ardecode:
> Hi! Thanks a lot for the great proposal! From my perspective, this is very strong and solid.
> 
> Personally, I have some questions and comments regarding this RFC, hope it could be some help.
> 
> 1. For metrics scraper, will you use the current one or a new one? Since the current one is designed to run in a longer interval, it might need to be modified a bit to have a better performance when scraping in a faster interval.
> 2. There is a failover mechanism which will re-route request to different endpoint, the lease might consider or be compatible with this scenario.
> 3. It would be great to mention the concurrency control while using the lock in the admission controller or the queue wake-up design.
> 4. I am wondering what will trigger the dequeue? Is it event-driven or is it gonna polling?
> 5. What will happen to the queued requests while shutting down the service?
> 
> Again, nice proposal :))

1. We plan to reuse the current EngineStatsScraper rather than introduce a second scraper service. The implementation will extend it with a queueing specific fast path when router queueing is enabled. This will allow us to keep the existing full scrape interval for the broader metric set, and add a faster admission oriented refresh focused on num_requests_waiting. That keeps backend metrics under one owner and avoids duplicated lifecycle management.

2. If I am not wrong, the current failover logic only reroutes when process_request() raises before the stream is established. So leases only need to be failover compatible for the pre first token failure path? The lease will be endpoint scoped. If a request fails before first token and the router decides to reroute, the implementation will release the old endpoint lease first and then acquire a fresh lease for the next endpoint. 

3. The admission controller will use a single asyncio.Lock to protect queue state and per-endpoint reservation counts. It will not use a wake-all design. Each queued request will wait on its own Future, and dequeue/drain will run under the lock, grant leases, update reservations, and resolve only the futures that were actually admitted. Routing and network I/O will happen outside the lock. 

4. The queue will be mainly event driven. We will try to dequeue when capacity changes locally, such as on first token lease release, pre first token error, cancellation, or timeout cleanup. In addition, the fast admission scrape will provide a lightweight periodic refresh so queued requests can still move forward when backend capacity changes independently of the router. So the design is event driven with a small periodic refresh, not pure polling.

5. On graceful shutdown, the router will stop admitting new queued requests, fail any requests still waiting in the queue, and release any undispatched leases. Requests that were already sent to a backend will follow the normal shutdown behavior. Since the queue is in memory, queued requests are not persisted across restart. We expect rejected queued requests to return a clear error, likely 503.

ruizhang0101:
I see. Could you add these to the RFC doc as well? Also, Could you also add a system diagram for this feature? And don't forget to sign-off the commit :).

PR #878, opened 2026-03-08
fix router memory #858

### LINKED ISSUE #858: Router default memory limit (500Mi) causes OOMKilled in real cluster deployments
## Problem

The current Helm chart version `vllm-stack` (e.g. 0.1.9) sets the default router memory limit to:

routerSpec:
  resources:
    limits:
      memory: 500Mi

In real cluster scenarios, this leads to:

- `OOMKilled`
- Exit code 137
- Router entering `CrashLoopBackOff`

500Mi is insufficient for typical production-stack usage.

---

## Observation

The GitHub repository appears to already contain an increased memory default for the router, but the published Helm chart (0.1.9) still renders `500Mi`.

Running:

helm show values vllm/vllm-stack --version 0.1.9

still shows `500Mi`.

So the Helm release does not yet include the updated default.

---

## Suggested Fix

Increase the default router memory limit to at least:

requests:
memory: 1Gi
limits:
memory: 2Gi

or remove the hard memory limit and leave it configurable.

---

## Workaround

Users can override the value manually in their `values.yaml`, but the default leads to immediate instability in many setups.

---

Thanks for the great project!

keyuchen21:
## Summary

The router memory limit was equal to the request (`1000Mi`), leaving zero
headroom for memory spikes from prefix cache growth and concurrent request
tracking. This caused OOMKilled crashes in production deployments.

## Changes
- Increase `routerSpec.resources.limits.memory` from `1000Mi` to `2000Mi`
- Fix stale README router resource defaults

## Test plan
- [ ] Deploy with default values and verify router pod no longer gets OOMKilled under load
- [ ] Verify `helm template` renders the correct memory limit (`2000Mi`)
- [ ] Confirm README table matches actual `values.yaml` defaults

Fixes #858

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -215,10 +215,10 @@ This table documents all available configuration values for the Production Stack
 | `routerSpec.requestStatsWindow` | integer | `60` | Window size in seconds for calculating request statistics |
 | `routerSpec.strategy` | map | `{}` | Deployment strategy for the router pods |
 | `routerSpec.vllmApiKey` | string/map | `null` | (Optional) API key for securing vLLM models |
-| `routerSpec.resources.requests.cpu` | string | `"4"` | CPU requests for router |
-| `routerSpec.resources.requests.memory` | string | `"16G"` | Memory requests for router |
-| `routerSpec.resources.limits.cpu` | string | `"8"` | CPU limits for router |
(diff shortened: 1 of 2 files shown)


keyuchen21:
will reopen the request later....currently a bit busy

Issue #861, opened 2026-03-02
cacheserverSpec missing from values.schema.json


cmhdave:
Working with the helm chart I noticed that cacheserverSpec and all associated settings are missing from values.schema.json.  This is strictly a documentation issue but noticed it because I fed the schema into an LLM to use for correctly formatting a values file for a helm install and it tripped up on the cacheserverSpec

ruizhang0101:
Thanks for pointing this out. We are working on a more robust way to structure the values.schema.json. It should be fixed soon.

Issue #751, opened 2025-11-05
feature:  allow custom labels in helm chart for Ray workerGroup


JustinDuy:
### Describe the feature

We have a small request to allow custom labels for Ray workerGroup. Our current situation is that we have a default istio side car injection to all pods, only can be disabled by specifying labels: `"sidecar.istio.io/inject: "false".` Unfortunately currently, there is no way to do that for the workerGroup: https://github.com/vllm-project/production-stack/blob/main/helm/templates/ray-cluster.yaml#L258

### Why do you need this feature?

vllm engine initialization inside ray head got stuck because the istio-proxy container inside ray worker intercepts all traffic. As a result, we can't start vllm engine with ray pipeline parallel.

### Additional context

Ray head custom label are passed via engineLabels: https://github.com/vllm-project/production-stack/blob/main/helm/templates/ray-cluster.yaml#L24 -> that should apply to workerGroup too in my opinion

JustinDuy:
Ray head log with 'peer reset connection to 172.31.226.15:44299' and istio-proxy log with that traffic <img width="1378" height="350" alt="Image" src="https://github.com/user-attachments/assets/a7077318-0c4e-4d6d-830d-204f3c5aabc0" />

JustinDuy:
@ahinsutime, @YuhanLiu11: if it's fine with you, i can add the code.

YuhanLiu11:
@JustinDuy sure! Please submit the PR with the added code. Thanks!

JustinDuy:
@YuhanLiu11 : could you please take a look at the MR? Thank you!

Issue #840, opened 2026-02-19
feature: Helm chart improvements


enneitex:
### Describe the feature

Hi,

While deploying the helm chart, I faced a few issues:
- PodDisruptionBudget does not work
- I can't deploy the serviceMonitor without the whole kube-prometheus-stack, prometheus operator is already deployed on my cluster
- there is no way to add extraManifest likean externalSecret
- I shouldn't need a bash sript to deploy the serviceMonitor and the grafana dashboard

### Why do you need this feature?

I think improving the helm chart would be nice so it can be "prod ready".
If the plan sounds good, I can submit a few PRs to
- address the PDB
- add a template for service Monitor directly in the vllm chart
- add a template for grafana dashboards
- add kube-prometheus-stack as a sub-chart
- add an extraManifest value

ruizhang0101:
Yes! I think it would be nice to add more k8s native configuration support for the current helm chart. :)

cmhdave:
While it's a little clunky, you can deploy ServiceMonitors using `extraObjects` which is how I'm doing it.  No extra kube-prometheus-stack or bash scripts

PR #886, opened 2026-03-14
[helm] document every values, udpate json schema and various fix

enneitex:
Contributes to https://github.com/vllm-project/production-stack/issues/840
fixes https://github.com/vllm-project/production-stack/issues/861
fixes https://github.com/vllm-project/production-stack/issues/751

Diff is huge but more than 2k LoC are from unit tests and 1700 lines for the json schema

## Changes

This PR is trying to fix a few things:

- Some values referenced in the `README.md` were wrong or missing from the `values.yaml`. 
- Some values from `values.yaml` were not used in the templates
- Some values in the template were missing in the `values.yaml`
- A lot of stuff was not working around `sharedPvcStorage`.There was no way to only create a PVC while a lot of people have access to storageClass and dynamic provisioning. PV is only created when specifying a nfs server or a hostPath
- `cacheserverSpec` and `sharedPvcStorage` were missing from the values so I added it with unit tests
- The json schema is now based on the `values.yaml` and updated by a pre commit hook
- Add an example for `modelSpec` and `loraAdapters`. they are not deployed by default but used to generate the schema

## To help review the PR

The json schema diff is hard to review. I ran `helm lint` with each value file from the CI to make sure they respect the updated schema

```bash
for f in ../.github/values*.yaml; do echo "=== $f ==="; helm lint . -f "$f"; done
for f in ../tutorials/assets/values*.yaml; do echo "=== $f ==="; helm lint . -f "$f"; done
```

I also ran `helm template` on each value file from `.github/values*.yaml` and `../tutorials/assets/values*.yaml` and compared the output between main and this branch. 

## Breaking changes

Before, this was enough to deploy the cache server
```yaml
cacheserverSpec:
  replicaCount: 1
```

Now it needs the following values:
```yaml
cacheserverSpec:
  enabled: true
  replicaCount: 1
```

Same thing for the rayClusters:

Before:
```yaml
modelSpec:
  - name: "model"
    raySpec:
      headNode:
        requestCPU: 1
```

After:
```yaml
modelSpec:
  - name: "model"
    raySpec:
      enabled: true
      headNode:
        requestCPU: 1
```

Even if the version is pre v1.0.0, a comment should be added to the release note and maybe a minor version bump (v0.2.0)

### DIFF
--- a/.github/values-09-kvaware-routing.yaml
+++ b/.github/values-09-kvaware-routing.yaml
@@ -21,9 +21,11 @@ servingEngineSpec:
         cpuOffloadingBufferSize: "10"
         enableController: true
         controllerPort: 9000
+        controllerReplyPort: 9001
         workerPorts: "8001"
         p2pHost: "localhost"
         p2pInitPorts: "30081"
+        workerHeartbeatTime: "30"
         logLevel: "DEBUG"
       env: []
       chatTemplate: "chat.jinja2"
@@ -57,3 +59,5 @@ routerSpec:
     - "--log-level"
     - "info"
   lmcacheControllerPort: 9000
+  lmcacheControllerReplyPort: 9001
+  lmcacheControllerHeartbeatPort: 9002
--- a/helm/templates/deployment-router.yaml
+++ b/helm/templates/deployment-router.yaml
@@ -136,6 +136,14 @@ spec:
           - "--lmcache-controller-port"
           - "{{ .Values.routerSpec.lmcacheControllerPort }}"
           {{- end }}
+          {{- if .Values.routerSpec.lmcacheControllerReplyPort }}
+          - "--lmcache-controller-reply-port"
+          - "{{ .Values.routerSpec.lmcacheControllerReplyPort }}"
+          {{- end }}
+          {{- if .Values.routerSpec.lmcacheControllerHeartbeatPort }}
+          - "--lmcache-controller-heartbeat-port"
+          - "{{ .Values.routerSpec.lmcacheControllerHeartbeatPort }}"
+          {{- end }}
           {{- if .Values.routerSpec.otel.endpoint }}
           - "--otel-endpoint"
           - "{{ .Values.routerSpec.otel.endpoint }}"
--- a/helm/templates/deployment-vllm-multi.yaml
+++ b/helm/templates/deployment-vllm-multi.yaml
@@ -356,6 +356,10 @@ spec:
           - name: LMCACHE_CONTROLLER_PULL_URL
             value: "{{ .Release.Name }}-router-service:{{ $modelSpec.lmcacheConfig.controllerPort }}"
           {{-   end }}
+          {{-   if hasKey $modelSpec.lmcacheConfig "controllerReplyPort" }}
+          - name: LMCACHE_CONTROLLER_REPLY_URL
+            value: "{{ .Release.Name }}-router-service:{{ $modelSpec.lmcacheConfig.controllerReplyPort }}"
+          {{-   end }}
           {{-   if hasKey $modelSpec.lmcacheConfig "workerPorts" }}
           - name: LMCACHE_LMCACHE_WORKER_PORTS
             value: {{ $modelSpec.lmcacheConfig.workerPorts | quote }}
--- a/helm/templates/ray-cluster.yaml
+++ b/helm/templates/ray-cluster.yaml
@@ -134,9 +134,17 @@ spec:
               - name: LMCACHE_CONTROLLER_URL
                 value: "{{ .Release.Name }}-{{$modelSpec.name}}-service:{{ $modelSpec.lmcacheConfig.controllerPort }}"
               {{-   end }}
+              {{-   if hasKey $modelSpec.lmcacheConfig "controllerReplyPort" }}
+              - name: LMCACHE_CONTROLLER_REPLY_URL
+                value: "{{ .Release.Name }}-{{$modelSpec.name}}-service:{{ $modelSpec.lmcacheConfig.controllerReplyPort }}"
+              {{-   end }}
               {{-   if hasKey $modelSpec.lmcacheConfig "workerPort" }}
               - name: LMCACHE_WORKER_PORT
                 value: "{{ .Release.Name }}-service:{{ $modelSpec.lmcacheConfig.workerPort }}"
+              {{-   end }}
+              {{-   if hasKey $modelSpec.lmcacheConfig "workerHeartbeatTime" }}
+              - name: LMCACHE_LMCACHE_WORKER_HEARTBEAT_TIME
+                value: {{ $modelSpec.lmcacheConfig.workerHeartbeatTime | quote }}
               {{-   end }}
               {{- end }}
             {{- if .Values.servingEngineSpec.configs }}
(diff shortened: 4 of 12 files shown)


helm/templates/deployment-cache-server.yaml
```diff
@@ -19,49 +27,50 @@ spec:
...
-      {{- if not (empty .Values.cacheserverSpec.affinity) }}
-      affinity:
-        {{- toYaml .Values.cacheserverSpec.affinity | nindent 8 }}
-      {{- else if .Values.cacheserverSpec.nodeSelectorTerms}}
+      {{- if .Values.cacheserverSpec.runtimeClassName }}
+      runtimeClassName: {{ .Values.cacheserverSpec.runtimeClassName }}
+      {{- end }}
+      {{- if .Values.cacheserverSpec.schedulerName }}
+      schedulerName: {{ .Values.cacheserverSpec.schedulerName }}
+      {{- end }}
```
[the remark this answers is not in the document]
enneitex: @gemini-code-assist The helm schema ensures these fields are strings. Looking at other popular charts, quote is not added to the template (https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/templates/argocd-server/deployment.yaml#L63 for example)

AlexanderSing:
Looking forward to the PR being merged, we also need the sharedPvcStorage fixes!

helm/templates/deployment-lora-adapter.yaml
```diff
@@ -9,16 +9,23 @@ metadata:
...
+  {{- else if and (kindIs "string" $.Values.servingEngineSpec.vllmApiKey) $.Values.servingEngineSpec.vllmApiKey }}
   vllmApiKey:
     secretName: {{ $.Release.Name }}-secrets
     secretKey: vllmApiKey
-  {{- else if .vllmApiKey }}
+  {{- else if and (kindIs "map" $.Values.servingEngineSpec.vllmApiKey) $.Values.servingEngineSpec.vllmApiKey.secretName $.Values.servingEngineSpec.vllmApiKey.secretKey }}
   vllmApiKey:
-    secretName: {{ .vllmApiKey.secretName | quote }}
-    secretKey: {{ .vllmApiKey.secretKey | quote }}
+    secretName: {{ $.Values.servingEngineSpec.vllmApiKey.secretName }}
```
ruizhang0101, line 27: Is it duplicated with the line 17-21?

tutorials/assets/keda-scaled-object.yaml
ruizhang0101, line 1: Is there a reason to change the file name?

enneitex: Yes, this file does not contain values so the name was misleading. 

The real goal behind this it to improve the CI step that runs `helm lint`. I would like to open a PR where helm lint is executed on every values files from tutorials and github workflows. Changing the name will allow me to do something like `grep values-*.yaml` without false positive.

helm/templates/deployment-lora-adapter.yaml
enneitex: I don't think so. There are three possible scenarios:

1.  Set the API key in the LoraAdapters. This only supports referencing a secret.
input is 
```yaml
loraAdapters:
  - name: "llama3-nemoguard-adapter"
    baseModel: "llama3"
    adapterSource:
      type: "local"
      adapterName: "llama-3.1-nemoguard-8b-topic-control"
      adapterPath: "/data/shared-pvc-storage/lora-adapters/llama-3.1-nemoguard-8b-topic-control"
    loraAdapterDeploymentConfig:
      algorithm: "default"
      replicas: 1
    vllmApiKey:
      secretName: apikey-secret
      secretKey: apikey
```
output is lines 17-20

2. Set the API key in servingEngineSpec as a string.
Input is 
```yaml
loraAdapters:
  - name: "llama3-nemoguard-adapter"
    baseModel: "llama3"
    adapterSource:
      type: "local"
      adapterName: "llama-3.1-nemoguard-8b-topic-control"
      adapterPath: "/data/shared-pvc-storage/lora-adapters/llama-3.1-nemoguard-8b-topic-control"
    loraAdapterDeploymentConfig:
      algorithm: "default"
      replicas: 1
servingEngineSpec:
  vllmApiKey: custom_api_key
```
output is lines 21-24. 

3. Set the API key in servingEngineSpec as a reference to an existing secret.
Input is 
```yaml
loraAdapters:
  - name: "llama3-nemoguard-adapter"
    baseModel: "llama3"
    adapterSource:
      type: "local"
      adapterName: "llama-3.1-nemoguard-8b-topic-control"
      adapterPath: "/data/shared-pvc-storage/lora-adapters/llama-3.1-nemoguard-8b-topic-control"
    loraAdapterDeploymentConfig:
      algorithm: "default"
      replicas: 1
servingEngineSpec:
  vllmApiKey:
    secretName: servingEngineSpec-apikey-secret
    secretKey: servingEngineSpec-apikey
```
output is lines 25-28. 
 
This is covered by unit tests in helm/tests/deployment-lora-adapter_test.yaml

It is tricky because vllmApiKey can be either a string or a map. It would be easier if the values were distinct fields, but this would be a breaking change.
```yaml
servingEngineSpec:
  vllmApiKey:
    apiKey: ""
    secretName: ""
    secretKey: ""
```

Exemple of values with distinct fields: https://github.com/bitnami/charts/blob/main/bitnami/postgresql/values.yaml#L41

tutorials/assets/keda-scaled-object.yaml
ruizhang0101: I see. Then would it be better to have some keda identification in the filename as well? Like keda-scale-object so that the ppl know it is related to keda.

enneitex: fixed in latest commit :)

ruizhang0101:
Could you resolve the conflict and then I think it is good to go :)

enneitex:
Hey @ruizhang0101 the PR has been rebased.
Some tests are failing but it does not look related to this PR

Reamer:
@enneitex 
With this PR, you are setting the image for the cache service to `lmcache/lmstack-cache-server`. This image does not exist on Docker Hub.
The old image was `lmcache/vllm-openai`. I think this should be changed.

enneitex:
Hi, 
I'll check this afternoon

enneitex:
Damn, I took the value from the README which mentioned `lmcache/lmstack-cache-server`, I'll open a small PR to update the values and readme

Reamer:
Sorry, I have to add one more thing.
You also set the following default for the cache service.
```
  containerSecurityContext:
    runAsNonRoot: false
```
In my opinion, you should either set this value to `true` or leave it out and let the K8s platform's Pod Security Standards handle it.

enneitex:
No problem!
I opened https://github.com/vllm-project/production-stack/pull/917

Reamer:
I'm back. With this PR, you added the `cacheserverSpec.livenessProbe` value to `values.yaml`. I'm now encountering the following issue, and perhaps you know why this behavior is occurring.

Problem: The file `/workspace/LMCache/examples/kubernetes/health_probe.py` is missing in releases of `lmcache/vllm-openai`, so the LivenessProbe naturally fails and the pod restarts continuously.

In `lmcache/vllm-openai:latest-nightly`, the file is included, but it is not in the releases `lmcache/vllm-openai:v0.4.3` and `lmcache/vllm-openai:v0.4.2`.

For now, I have removed the livenessProbe. It’s not ideal, but I wanted to let you know about it.

PR #897, opened 2026-03-26
[Feat] Helm: add support for per-model tolerations

AlexanderSing:
## Summary

- Adds support for per-model `tolerations` in `modelSpec`, allowing individual model deployments to have their own Kubernetes tolerations in addition to or instead of the global `servingEngineSpec.tolerations`
- Introduces a `tolerationsPolicy` field per model spec with two modes: `append` (default) merges global + model tolerations, and `override` replaces global tolerations entirely for that model
- Updates `values.schema.json`, `values.yaml`, and `values-example.yaml` with documentation and schema definitions for the new fields
- Adds a comprehensive Helm unit test suite (`tolerations_test.yaml`) covering all policy combinations and edge cases

## Motivation

In multi-model deployments, different models may run on different node pools with distinct taints. Previously, tolerations could only be set globally for all models. This change enables operators to schedule models on specific node types (e.g., different GPU tiers) without requiring separate Helm releases.

## Behavior

| Scenario | Result |
|---|---|
| No model tolerations set | Global tolerations applied |
| Model tolerations + `append` policy (default) | Global + model tolerations merged |
| Model tolerations + `override` policy | Model tolerations only (global ignored) |
| Model tolerations, no global tolerations | Model tolerations applied |
| No tolerations anywhere | No `tolerations` field rendered |


helm/values.yaml
```diff
@@ -120,9 +120,19 @@ servingEngineSpec:
   #
   # - affinity: (optional, map) Affinity configuration
   # - nodeSelectorTerms: (optional, list) The node selector terms to match the nodes. When both affinity and nodeSelectorTerms are defined, nodeSelectorTerms will be ignored.
-  # - nodeName: (optional) Directly assigns a pod to a specific node (e.g., "192.168.56.5"). When both nodeName and nodeSelectorTerms are defined, the preference is given to nodeName.
-  # - shmSize: (optional, string) The size of the shared memory, e.g., "20Gi"
-  # - enableLoRA: (optional, bool) Whether to enable LoRA, e.g., true
+   # - nodeName: (optional) Directly assigns a pod to a specific node (e.g., "192.168.56.5"). When both nodeName and nodeSelectorTerms are defined, the preference is given to nodeName.
```
ruizhang0101: The indention seems a little bit off. Could you align with the others?

helm/values.schema.json
```diff
@@ -506,25 +506,57 @@
...
-              "requestCPU",
-              "requestMemory",
-              "requestGPU",
-              "pvcStorage"
-            ]
-          }
-        },
-        "containerPort": {
+                   "required": [
+                   "enabled"
```
ruizhang0101: The indention seems a bit off.

AlexanderSing:
> Thanks for the PR! The ray deployment (ray-cluster.yaml) is also using tolerations. Would it worth changing as well? Also, could you add this to helm readme too?

Thank you for your review, really appreciate it a lot! I think i fixed all the indentation issues, added the README entries and also added the per model tolerations to the ray-cluster!

Open to further suggestions and feedback!

ruizhang0101:
Could you resolve this conflicts? Then I think it is good to go :))

AlexanderSing:
Hi i resolved it - I guess we're waiting on the fixes for the issues from the other recently merged PR right? :)

ruizhang0101:
Yes. Also could you fix the pre-commit issue when you have some time?

PR #903, opened 2026-03-30
[Feat] Support KEDA Auto Scaling in Production Stack Operator 

aeon-x:
## Summary

Add built-in KEDA autoscaling support so that the operator automatically creates a ScaledObject for each VLLMRuntime CR when autoscaling is enabled. This enables per-deployment autoscaling (including scale-to-zero) based on Prometheus metrics without manually creating ScaledObjects.

 ### Key design decision
 KEDA targets the VLLMRuntime CR via its scale subresource (not the Deployment directly), so the CR remains the single source of truth for replicas. This eliminates the operator/KEDA replica conflict: KEDA updates CR replicas → operator reconciles → Deployment updated.

  ## Changes

  ### CRD (`api/v1alpha1/vllmruntime_types.go`)
  - Added scale subresource marker (`specpath`, `statuspath`, `selectorpath`) to VLLMRuntime
  - Added `Replicas` and `Selector` fields to `VLLMRuntimeStatus` (required by HPA for `AverageValue` metric division)
  - Added `AutoscalingConfig` struct with configurable thresholds, timing, and HPA stabilization windows

  ### Controller (`internal/controller/vllmruntime_controller.go`)
  - Added `reconcileScaledObject()` — creates/updates KEDA ScaledObject via server-side apply, gated by `AutoscalingConfig.Enabled`
  - Updated `updateStatus()` to set `status.replicas` and `status.selector` from the Deployment
  - Added RBAC markers for `keda.sh/scaledobjects` and `vllmruntimes/scale`

  ### Prometheus triggers (4 total)
  1. **Router keepalive** (`metricType: Value`) — controls 0↔1 scaling, always available even with 0 pods
  2. **Concurrent requests** (`AverageValue`) — `sum(vllm:num_requests_running)`, threshold: 5/pod
  3. **Generation tokens/s** (`AverageValue`) — `sum(rate(vllm:generation_tokens_total[1m]))`, threshold: 100/pod
  4. **Prompt tokens/s** (`AverageValue`) — `sum(rate(vllm:prompt_tokens_total[1m]))`, threshold: 100/pod

  ### Generated manifests
  - Regenerated CRD with scale subresource and `autoscalingConfig` schema
  - Regenerated RBAC with `keda.sh` and `/scale` permissions

  ## Example CR

  ```yaml
  apiVersion: production-stack.vllm.ai/v1alpha1
  kind: VLLMRuntime
  spec:
    autoscalingConfig:
      enabled: true
      minReplicas: 0
      maxReplicas: 4
      requestsRunningThreshold: "5"
      generationTokensThreshold: "100"
      promptTokensThreshold: "100"
      pollingInterval: 15
      cooldownPeriod: 1800
      scaleUpStabilizationWindowSeconds: 0
      scaleDownStabilizationWindowSeconds: 300


operator/api/v1alpha1/vllmruntime_types.go
```diff
@@ -51,6 +51,43 @@ type DeploymentConfig struct {
...
+// AutoscalingConfig defines the KEDA autoscaling configuration
+type AutoscalingConfig struct {
+	// Enabled enables autoscaling
+	Enabled bool `json:"enabled,omitempty"`
+
+	// MinReplicas is the minimum number of replicas (0 enables scale-to-zero)
+	MinReplicas int32 `json:"minReplicas"`
+
+	// MaxReplicas is the maximum number of replicas (if not set, defaults to MinReplicas)
+	MaxReplicas *int32 `json:"maxReplicas,omitempty"`
```
ruizhang0101: Will there will be scenario that if the min replica is 0, and max replica not set. So max replica defaults to 0 as well. Then the object will never be scaled? 
Could you also add some validation of max and min replicas?

operator/internal/controller/vllmruntime_controller.go
```diff
@@ -1030,10 +1051,174 @@ func (r *VLLMRuntimeReconciler) updateStatus(
...
+		"maxReplicaCount": getMaxReplicas(cfg),
+		"pollingInterval": getPollingInterval(cfg),
+		"cooldownPeriod":  getCooldownPeriod(cfg),
+		"advanced": map[string]interface{}{
+			"horizontalPodAutoscalerConfig": map[string]interface{}{
+				"behavior": map[string]interface{}{
+					"scaleUp": map[string]interface{}{
+						"stabilizationWindowSeconds": getScaleUpWindow(cfg),
+						"policies": []map[string]interface{}{
+							{"type": "Pods", "value": 1, "periodSeconds": 60},
```
ruizhang0101: Can scale up/down policy be configurable with some default value?

operator/internal/controller/vllmruntime_controller.go
```diff
@@ -1030,10 +1051,174 @@ func (r *VLLMRuntimeReconciler) updateStatus(
...
+			},
+		},
+		"triggers": []map[string]interface{}{
+			{
+				"type":       "prometheus",
+				"metricType": "Value",
+				"metadata": map[string]string{
+					"serverAddress": prometheusAddr,
+					"metricName":    "vllm_incoming_keepalive",
+					"query":         fmt.Sprintf(`sum(rate(vllm:num_incoming_requests_total{namespace="%s", model="%s"}[1m]) > bool 0)`, vllmRuntime.Namespace, servedModelName),
```
ruizhang0101, line 1128: is it possible to use job label as well to align with the other triggers?

operator/internal/controller/vllmruntime_controller.go
```diff
@@ -1030,10 +1051,174 @@ func (r *VLLMRuntimeReconciler) updateStatus(
...
+}
+
+func getMaxReplicas(cfg *productionstackv1alpha1.AutoscalingConfig) int32 {
+	if cfg.MaxReplicas != nil {
+		return *cfg.MaxReplicas
+	}
+	return cfg.MinReplicas
+}
+
+func getPollingInterval(cfg *productionstackv1alpha1.AutoscalingConfig) int32 {
```
ruizhang0101: Could you use kubebuilder default value for these kind of static value?

operator/internal/controller/vllmruntime_controller.go
aeon-x: The keepalive trigger uses namespace + model labels because it comes from the router, not the vLLM pod. The router is always running (even when pods are scaled to zero), which is why it can trigger scale-from-zero.

The other triggers use job label because they come from the vLLM pod metrics — those are only available when pods are running.

operator/api/v1alpha1/vllmruntime_types.go
aeon-x: Yes, updated.

operator/internal/controller/vllmruntime_controller.go
aeon-x: Updated

operator/internal/controller/vllmruntime_controller.go
aeon-x: updated. removed getters/

PR #916, opened 2026-04-14
[Bugfix] fix(vllm-router): keep roundrobin state per endpoint set / model

### LINKED ISSUE #915: bug: roundrobin routing is global instead of per endpoint set / model
`RoundRobinRouter` uses one global counter for all requests, not one counter per endpoint set / model.

In multi-model deployments, traffic for other models shifts the round-robin state, so a model with 2 backends can get uneven distribution even with `routing_logic: roundrobin`.

Expected: round robin should be fair within the requested endpoint set / model.

Suggested fix: track round-robin state per endpoint set / model (for example by endpoint URL hash/tuple).

Relevant code:
- `src/vllm_router/routers/routing_logic.py`
- `src/vllm_router/services/request_service/request.py`

Killusions:
Keeps round-robin state per endpoint set / model instead of sharing one global counter across all traffic.

Adds regression coverage for interleaved endpoint sets and reordered endpoint lists.

Closes #915

⚒️ with ❤️ by @siemens

### DIFF
--- a/src/vllm_router/routers/routing_logic.py
+++ b/src/vllm_router/routers/routing_logic.py
@@ -59,7 +59,6 @@ class RoutingLogic(str, enum.Enum):
 
 
 class RoutingInterface(metaclass=SingletonABCMeta):
-
     def _qps_routing(
         self, endpoints: List[EndpointInfo], request_stats: Dict[str, RequestStats]
     ) -> str:
@@ -140,14 +139,31 @@ class RoutingInterface(metaclass=SingletonABCMeta):
 class RoundRobinRouter(RoutingInterface):
     # TODO (ApostaC): when available engines in the endpoints changes, the
     # algorithm may not be "perfectly" round-robin.
+
+    # Upper bound on cached endpoint-set entries to prevent unbounded memory
+    # growth when endpoints change dynamically (add / remove / update).
+    _MAX_CACHE_SIZE = 1024
+
     def __init__(self):
         if hasattr(self, "_initialized"):
             return
-        self.req_id = 0
-        self.sorted_endpoints = []
-        self.last_endpoints_id = None
-        self.last_endpoints_hash = None
+        self._next_index: dict[tuple[str, ...], int] = {}
+        self._sorted_cache: dict[frozenset[str], tuple[str, ...]] = {}
         self._initialized = True
+
+    def _endpoint_key(self, endpoints: List[EndpointInfo]) -> tuple[str, ...]:
+        """Return a stable, sorted key for the endpoint set (cached after first sort)."""
+        if not endpoints:
+            raise ValueError("RoundRobinRouter requires at least one endpoint")
+
+        urls = frozenset(e.url for e in endpoints)
+        key = self._sorted_cache.get(urls)
+        if key is None:
+            if len(self._sorted_cache) >= self._MAX_CACHE_SIZE:
+                self._sorted_cache.clear()
+            key = tuple(sorted(urls))
+            self._sorted_cache[urls] = key
+        return key
 
     def route_request(
         self,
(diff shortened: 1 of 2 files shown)


src/vllm_router/routers/routing_logic.py
```diff
@@ -143,10 +142,8 @@ class RoundRobinRouter(RoutingInterface):
     def __init__(self):
         if hasattr(self, "_initialized"):
             return
-        self.req_id = 0
-        self.sorted_endpoints = []
-        self.last_endpoints_id = None
-        self.last_endpoints_hash = None
+        self.endpoint_sets = {}
```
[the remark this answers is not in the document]
Killusions: This is not a real problem, no one has that many endpoints.

src/vllm_router/routers/routing_logic.py
```diff
@@ -143,10 +142,8 @@ class RoundRobinRouter(RoutingInterface):
...
+        self.req_ids = {}
```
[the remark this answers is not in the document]
Killusions: This is theoretical, it's realistically only used once.

src/vllm_router/routers/routing_logic.py
```diff
@@ -168,16 +165,15 @@ def route_request(
...
-        return chosen.url
+        endpoint_set = frozenset(endpoint.url for endpoint in endpoints)
+        endpoint_urls = self.endpoint_sets.get(endpoint_set)
+        if endpoint_urls is None:
+            endpoint_urls = tuple(sorted(endpoint_set))
+            self.endpoint_sets[endpoint_set] = endpoint_urls
+        req_id = self.req_ids.get(endpoint_urls, 0)
+        chosen = endpoint_urls[req_id % len(endpoint_urls)]
+        self.req_ids[endpoint_urls] = req_id + 1
+        return chosen
```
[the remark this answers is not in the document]
Killusions: 1. Should be fine in the actual implementation, 2. fixed.

Killusions:
@ruizhang0101 Could you maybe take a look at this?

Killusions:
@ruizhang0101 Any update on this? We've had to switch routing strategies because of this.

src/vllm_router/routers/routing_logic.py
```diff
@@ -143,12 +142,16 @@ class RoundRobinRouter(RoutingInterface):
...
-        self.last_endpoints_hash = None
+        self.next_index_by_endpoint_set = {}
         self._initialized = True
 
+    def _get_endpoint_urls(self, endpoints: List[EndpointInfo]) -> tuple[str, ...]:
+        if not endpoints:
+            raise ValueError("RoundRobinRouter requires at least one endpoint")
+
+        # Treat the same endpoint set as one rotation, even if discovery reorders it.
+        return tuple(sorted(endpoint.url for endpoint in endpoints))
```
ruizhang0101: Could you keep id and hash methods to avoid sorting for every request?

Killusions: @ruizhang0101 Do you prefer something like this?

ruizhang0101: @Killusions LGTM. Other than that, I am little concerned about the unbounded memory of these two dicts _next_index and sorted_cache.

Killusions: @ruizhang0101 That's fair, but I guess the upper bound is the number of defined endpoint sets (= models), as it should not be anywhere close to a real memory leak. Unless I'm misunderstanding something

ruizhang0101: @Killusions It is true under static setting. Under dynamic setting, whenever the endpoints of the model change, either update, add or delete, it will create an entry in the dict. If I understand this correctly.

Killusions: @ruizhang0101 I guess that's true, will add an upper bound of like 1k

Killusions: @ruizhang0101 Added upper bounds and fixed pre-commit.

ruizhang0101 (dismissed):
LGTM

Killusions:
@ruizhang0101 Thanks, fixed the one sorting issue again, but the other failing pipelines are because of the fork I guess

ruizhang0101 (approved):
LGTM

PR #917, opened 2026-04-14
fix(helm) fix default values for cache deployment

enneitex:
Contributes to https://github.com/vllm-project/production-stack/issues/840

This is a follow-up to https://github.com/vllm-project/production-stack/pull/886.

It fixes two default values for the cache server.


Reamer:
LGTM

Reamer (approved):

ruizhang0101 (approved):
LGTM

PR #918, opened 2026-04-15
[Minor Improvements] Vllmruntime Autoscaling in Operator

aeon-x:
## Summary

  ### Three improvements to VLLMRuntime KEDA autoscaling:

  1. Fix: resolve --served-model-name override in Prometheus queries — The keepalive trigger queries Prometheus with model="<name>". Previously it always used ModelURL (e.g., meta-llama/Llama-3.1-8B-Instruct), ignoring --served-model-name in ExtraArgs. If a user overrides the served model name, the query returns no data and KEDA may incorrectly scale to zero while traffic is active.
  2. Widen keepalive rate window from [1m] to [2m] — A 1-minute window is tight; delayed Prometheus scrapes or brief metric gaps can return 0 and trigger a false scale-to-zero. A 2-minute window is more resilient.
  3. Add AvailableReplicas, UpdatedReplicas, UnavailableReplicas to VLLMRuntimeStatus — Exposes deployment replica status directly on the CR, so users can observe scaling progress via kubectl get vr without inspecting the underlying Deployment.

  ### Changes:

  - operator/api/v1alpha1/vllmruntime_types.go — Add 3 status fields
  - operator/internal/controller/vllmruntime_controller.go — Served-model-name parsing, 2m keepalive window, populate new status fields
  - operator/api/v1alpha1/zz_generated.deepcopy.go — Regenerated
  - operator/config/crd/bases/production-stack.vllm.ai_vllmruntimes.yaml — Regenerated

  ### Tests Done: 

  - make generate manifests — regenerated without errors
  - make test — all VLLMRuntime and autoscaling tests pass (11/11)
  - Deploy with a VLLMRuntime CR using --served-model-name override and verify KEDA ScaledObject uses the correct model name in Prometheus queries
  - Verify kubectl get vr shows the new replica status fields

### DIFF
--- a/operator/api/v1alpha1/vllmruntime_types.go
+++ b/operator/api/v1alpha1/vllmruntime_types.go
@@ -328,6 +328,15 @@ type VLLMRuntimeStatus struct {
 	// Current replica count (used by scale subresource)
 	Replicas int32 `json:"replicas,omitempty"`
 
+	// Number of pods that are ready and serving traffic
+	AvailableReplicas int32 `json:"availableReplicas,omitempty"`
+
+	// Number of pods running the latest pod template
+	UpdatedReplicas int32 `json:"updatedReplicas,omitempty"`
+
+	// Number of pods that are not yet available
+	UnavailableReplicas int32 `json:"unavailableReplicas,omitempty"`
+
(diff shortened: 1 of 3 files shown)


ruizhang0101 (approved):
LGTM

PR #936, opened 2026-04-29
[Bugfix][Router] Fix stale Prometheus endpoint health metrics (#888)

### LINKED ISSUE #888: bug: Router - Stale Prometheus endpoint health metrics when engines become unavailable in service discovery
## Summary
When a discovered backend becomes unhealthy/unready, it is removed from service discovery, but Prometheus metrics can continue exposing the old value for that endpoint label.

## Problem
In service discovery, engines are [removed](https://github.com/vllm-project/production-stack/blob/e92d43fb3dff3372a2b67bc653e9ad6283d3afc4/src/vllm_router/service_discovery.py#L746-L749) from `available_engines` when they are no longer ready (for example in `MODIFIED` events with `not is_pod_ready` / `not is_service_ready` or empty model list).

After removal, that endpoint no longer appears in `get_endpoint_info()`, so the [metrics loop](https://github.com/vllm-project/production-stack/blob/e92d43fb3dff3372a2b67bc653e9ad6283d3afc4/src/vllm_router/routers/metrics_router.py#L115-L120) does not overwrite its previous gauge value.

As a result, stale series remain exported and dashboards/alerts can show outdated health state.

## Additional issue
`EndpointInfo` does not model a `healthy` field, while metrics currently rely on `getattr(ep, "healthy", True)`.
That fallback treats missing health as healthy, which can hide real state transitions.

### To Reproduce

1. Start with one healthy endpoint exposed in metrics (`healthy_pods_total{server=...}=1`).
2. Make that backend unready/unhealthy so service discovery removes it from `available_engines`.
3. Scrape metrics again.
4. Observe that old label values remains unless gauge labels are explicitly cleared/reset.

### Expected behavior

- Exported metrics should only represent currently discovered endpoints.
- Removed endpoints should not keep stale values.
- Endpoint health should be explicit and strongly typed, not inferred with fallback defaults.

## Proposed fix
In [metrics_router.py](https://github.com/vllm-project/production-stack/blob/main/src/vllm_router/routers/metrics_router.py)
1. Clear label-based gauges before repopulating metrics.
```

def _clear_label_gauges() -> None:
    # Clear all label-based gauges to avoid stale series for removed endpoints.
    healthy_pods_total.clear()
    current_qps.clear()
    avg_decoding_length.clear()
    num_prefill_requests.clear()
    num_decoding_requests.clear()
    num_requests_running.clear()
    avg_latency.clear()
    avg_itl.clear()
    num_requests_swapped.clear()
    gpu_prefix_cache_hit_rate.clear()
    gpu_prefix_cache_hits_total.clear()
    gpu_prefix_cache_queries_total.clear()


@metrics_router.get("/metrics")
async def metrics():

    _clear_label_gauges()
    ...
```
2. Add `healthy: bool` to `EndpointInfo` and set it in service discovery implementations.
3. Remove fallback `getattr(..., True)` and read explicit health state.

### Some non-LLM thoughts
I tested this myself in a K8s setup, and it works in terms of showing updated/non-stale metrics. We rely on such metrics in an upstream service to understand whether the underlying nodes can handle more traffic.

I’m not sure whether this fix/approach affects other features, so I held off on opening a PR. If you like this solution, I can open one with the proposed changes.

prashansapkota:
## Summary

Fixes #888

Endpoint health gauges retained stale `server=` labels indefinitely when a backend was removed from service discovery, causing Prometheus to keep exporting the last known value for dead endpoints.

- Add `EndpointInfo.healthy` field (default `True`) to carry health state from service discovery into the metrics handler
- Introduce `_LABEL_GAUGES` list and `_clear_label_gauges()` helper in `metrics_router.py`; call it at the start of every `/metrics` scrape so only currently-active endpoints appear in the output
- Update `healthy_pods_total` to use `ep.healthy` instead of a hardcoded `1`
- Add `test_stale_metrics.py` with 7 tests covering: `EndpointInfo.healthy` default/override, stale label removal, full endpoint removal, unlabeled gauge preservation, `_LABEL_GAUGES` completeness, and post-clear repopulation

## Test plan

- `pytest src/tests/test_stale_metrics.py`. All 7 tests pass
- `pytest src/tests/test_static_service_discovery.py src/tests/test_utils.py`. All existing tests pass
-  Manually verify `/metrics` no longer exports labels for removed endpoints after a scrape cycle

### DIFF
--- a/src/tests/test_stale_metrics.py
+++ b/src/tests/test_stale_metrics.py
@@ -1,0 +1,117 @@
+import pytest
+from prometheus_client import REGISTRY, CollectorRegistry, generate_latest
+
+from vllm_router.routers.metrics_router import _LABEL_GAUGES, _clear_label_gauges
+from vllm_router.service_discovery import EndpointInfo
+from vllm_router.services.metrics_service import (
+    current_qps,
+    gpu_prefix_cache_hit_rate,
+    gpu_prefix_cache_hits_total,
+    gpu_prefix_cache_queries_total,
+    healthy_pods_total,
+    num_requests_running,
+)
+
+
+def test_endpoint_info_healthy_defaults_to_true():
+    ep = EndpointInfo(
+        url="http://ep1:8000",
+        model_names=["llama"],
+        Id="id1",
+        added_timestamp=0,
+        model_label="default",
+        sleep=False,
+    )
+    assert ep.healthy is True
+
+
+def test_endpoint_info_healthy_can_be_set_false():
+    ep = EndpointInfo(
+        url="http://ep1:8000",
+        model_names=["llama"],
+        Id="id1",
+        added_timestamp=0,
+        model_label="default",
+        sleep=False,
+        healthy=False,
+    )
+    assert ep.healthy is False
+
+
+@pytest.fixture(autouse=True)
+def _reset_gauges():
+    """Clear every label-based gauge before and after each test."""
+    _clear_label_gauges()
+    yield
+    _clear_label_gauges()
+
+
+def test_cleared_gauge_removes_stale_labels():
+    """Stale server labels must disappear after _clear_label_gauges()."""
+    # Simulate two active endpoints
+    healthy_pods_total.labels(server="http://ep1:8000").set(1)
+    healthy_pods_total.labels(server="http://ep2:8000").set(1)
+
+    output = generate_latest(REGISTRY).decode()
+    assert "ep1" in output
+    assert "ep2" in output
+
+    # Clear all labels (simulates start of /metrics handler)
+    _clear_label_gauges()
+
+    # Re-populate only ep1 (ep2 was removed from service discovery)
+    healthy_pods_total.labels(server="http://ep1:8000").set(1)
+
+    output = generate_latest(REGISTRY).decode()
+    assert "ep1" in output
+    assert "ep2" not in output
+
+
+def test_clear_removes_all_labels_when_all_endpoints_gone():
+    """When every endpoint is removed, no server labels remain."""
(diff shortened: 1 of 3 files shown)


prashansapkota:
Thanks for the review! Addressed all feedback in the latest commit (36748d0):

1. **Added `num_requests_waiting` to `_LABEL_GAUGES`**: this gauge has a `server` label and was missing from the clearing list, meaning stale waiting-request series would have persisted for removed endpoints. Also added it to the import in `metrics_router.py`.

2. **Restored the `metrics()` endpoint docstring**: the docstring was removed as part of a comment cleanup pass but reviewers flagged it as valuable for maintainability. It's been restored in full.

3. **Expanded imports in `test_stale_metrics.py`**: the test file was only importing a subset of server-labeled gauges. All gauges (`avg_decoding_length`, `avg_itl`, `avg_latency`, `num_decoding_requests`, `num_prefill_requests`, `num_requests_swapped`, `num_requests_waiting`) are now imported to allow comprehensive coverage.

4. **Expanded `expected` set in `test_label_gauges_list_contains_all_expected_gauges`**: the set now mirrors the full `_LABEL_GAUGES` list exactly, including `num_requests_waiting`, ensuring the test catches any future omissions from the clearing list.

ruizhang0101 (dismissed):
LGTM

ruizhang0101:
@prashansapkota Hi, Could you fix the pre-commit when you have time?

prashansapkota:
@ruizhang0101 Hi, Fixed the pre-commit failure, removed the unused `CollectorRegistry` import in `test_stale_metrics.py` that ruff was flagging. All pre-commit checks pass now.

ruizhang0101 (approved):
LGTM

PR #939, opened 2026-05-01
[Feat][Router]: Add automatic retry with exponential backoff and jitter

ikaadil:
## Overview

Adds opt-in automatic retry for transient backend failures using exponential backoff with jitter, and folds the existing instance-failover reroute into the same mechanism.

Motivation: [Mitigating the thundering herd problem](https://medium.com/@avnein4988/mitigating-the-thundering-herd-problem-exponential-backoff-with-jitter-b507cdf90d62)

A forwarded request fails in one of two ways, which call for opposite responses:

| Failure | Example | Response |
| --- | --- | --- |
| Transport failure | connection refused, timeout | Engine excluded, request rerouted to another engine **immediately** |
| Retryable status | `408` `429` `500` `502` `503` `504` | Engine stays eligible, request re-issued **after a backoff** |

Both draw on one budget: `--max-retries`. Retrying is **disabled by default**, so a request is attempted exactly once unless `--enable-retries` is passed.

## CLI

| Flag | Default | Description |
| --- | --- | --- |
| `--enable-retries` | off | Enable retrying |
| `--max-retries` | `5` | Total attempts **including the initial one**; must be >= 2 |
| `--initial-backoff-ms` | `50` | Backoff before the first retry |
| `--max-backoff-ms` | `30000` | Upper bound on the backoff |
| `--backoff-multiplier` | `1.5` | Growth factor between retries |
| `--jitter-factor` | `0.2` | Randomisation applied to each delay (0.0-1.0) |

Without `--enable-retries` the other flags are ignored, so a shared config template carrying them cannot change behaviour or block startup.

```bash
vllm-router --port 8000 \
    --service-discovery static \
    --static-backends "http://localhost:9001,http://localhost:9002" \
    --static-models "facebook/opt-125m,facebook/opt-125m" \
    --routing-logic roundrobin \
    --enable-retries \
    --max-retries 5 \
    --initial-backoff-ms 100 \
    --max-backoff-ms 60000 \
    --backoff-multiplier 2.0 \
    --jitter-factor 0.1
```

## Backoff

```text
delay  = min(initial_backoff_ms * multiplier ^ retry, max_backoff_ms)
delay' = delay * (1 + U[-jitter_factor, +jitter_factor])
```

With the defaults, retries land at `50ms`, `75ms`, `112.5ms`, `168.8ms`, each spread across a +/-20% window. Without the jitter, routers that backed off from the same incident retry in lockstep and re-create the overload they backed off from.

## Changes

| File | Change |
| --- | --- |
| `services/request_service/retry.py` | **New.** `RetryConfig` (frozen, self-validating), `RetryState` (attempt sequencing), `is_retryable_status` |
| `services/request_service/request.py` | `route_general_request` drives `RetryState.attempts()`; shared router dispatch extracted into `_select_backend()` |
| `parsers/parser.py` | Retry flags; validation delegated to `RetryConfig` |
| `app.py` | `app.state.retry_config = RetryConfig.from_args(args)` |
| `routers/routing_logic.py` | Retry policy moved out; failover wiring removed |

Validation lives in `RetryConfig.__post_init__`, so an invalid config cannot be constructed.

## Behaviour

- A retryable status does not blacklist the engine, so it stays a candidate after the backoff. This is what makes retrying useful on a single-engine deployment.
- Failover between healthy engines is never delayed; the backoff applies only once every engine has been excluded and the pool is given another chance.
- Once the budget is spent the backend response passes through, so the client sees the engine's own status and body rather than a synthesised router error.
- Non-retryable statuses return on the first attempt.
- Retries apply only before streaming begins; once the first chunk is forwarded the headers are already with the client, so a mid-stream error passes through untouched.

## Breaking change

`--max-instance-failover-reroute-attempts` is **removed**. It rerouted a failed request to another engine and is now subsumed by `--max-retries`, which covers the same case and adds backoff.

Its default was `0` (a single attempt), which is also the default here, so deployments that never set it are unaffected. Deployments that did set it should migrate:

```diff
- --max-instance-failover-reroute-attempts 2
+ --enable-retries --max-retries 3
```

## Note for reviewers

`process_request` reports a backend error status by **yielding** it (`yield backend_response.headers, backend_response.status`), not by raising, and there is no `raise_for_status` on that call. A retry implementation that only catches `HTTPException` never fires for an engine 5xx. This PR inspects the status after the first `anext`, and `aclose()`s a discarded response so the upstream connection is released.

## Tests

56 new tests, 281 passing suite-wide.

- `test_retry_config.py` — status classification, backoff growth, cap, jitter bounds, validation, `from_args` mapping.
- `test_request_retry.py` — end to end through `route_general_request`: default single attempt, transport reroute, no-backoff-on-failover, single-engine retry, budget exhaustion passthrough, non-retryable passthrough, generator close on discard, exact sleep sequence.

Every backend-status stand-in yields the status, matching `process_request`; one that raises `HTTPException` would pass whether or not retrying works.


ikaadil:
@ruizhang0101 could you please review the MR? Thanks!

ruizhang0101:
@aeon-x Could you take a look at this?

aeon-x:
Hey @ikaadil, i think retrying should be an optional, as most users would need fast fail over. 

Can you make sure that this retry mechanism is turned off unless it is explictly turned on by a flag?

ikaadil:
> Can you make sure that this retry mechanism is turned off unless it is explictly turned on by a flag?

Done

src/vllm_router/README.md
```diff
@@ -61,6 +61,41 @@ The router can be configured using command-line arguments. Below are the availab
...
+- Buffering entire responses would defeat the purpose of streaming
+- HTTP headers have already been sent to the client
+- Clients can implement their own retry logic for incomplete streaming responses
+
+- `--enable-retries`: Enable automatic retry for transient HTTP failures. Disabled by default.
+- `--retry-max-retries`: Maximum total attempts including the initial request. Default is 5 (meaning 1 initial attempt + up to 4 retries). Only used when `--enable-retries` is set.
+- `--retry-initial-backoff-ms`: Initial backoff duration in milliseconds. Default is 50.
+- `--retry-max-backoff-ms`: Maximum backoff duration in milliseconds. Default is 30000.
+- `--retry-backoff-multiplier`: Exponential backoff multiplier. Default is 1.5.
+- `--retry-jitter-factor`: Random jitter factor (0.0-1.0) to prevent thundering herd. Default is 0.2.
```
waelrabah11:
```suggestion
- `--max-retries`: Maximum total attempts including the initial request. Default is 5 (meaning 1 initial attempt + up to 4 retries). Only used when `--enable-retries` is set.
- `--initial-backoff-ms`: Initial backoff duration in milliseconds. Default is 50.
- `--max-backoff-ms`: Maximum backoff duration in milliseconds. Default is 30000.
- `--backoff-multiplier`: Exponential backoff multiplier. Default is 1.5.
- `--jitter-factor`: Random jitter factor (0.0-1.0) to prevent thundering herd. Default is 0.2.
```

src/vllm_router/README.md
```diff
@@ -61,6 +61,41 @@ The router can be configured using command-line arguments. Below are the availab
...
+    --service-discovery static \
+    --static-backends "http://localhost:9001,http://localhost:9002" \
+    --static-models "facebook/opt-125m,facebook/opt-125m" \
+    --routing-logic roundrobin \
+    --enable-retries \
+    --retry-max-retries 5 \
+    --retry-initial-backoff-ms 100 \
+    --retry-max-backoff-ms 60000 \
+    --retry-backoff-multiplier 2.0 \
+    --retry-jitter-factor 0.1
```
waelrabah11:
```suggestion
    --max-retries 5 \
    --initial-backoff-ms 100 \
    --max-backoff-ms 60000 \
    --backoff-multiplier 2.0 \
    --jitter-factor 0.1
```

src/vllm_router/app.py
```diff
@@ -270,6 +271,19 @@ def initialize_all(app: FastAPI, args):
...
         configure_custom_callbacks(args.callbacks, app)
 
+    # Configure retry mechanism (disabled by default for fast failover)
+    if args.enable_retries:
+        app.state.retry_config = RetryConfig(
+            max_retries=args.retry_max_retries,
+            initial_backoff_ms=args.retry_initial_backoff_ms,
+            max_backoff_ms=args.retry_max_backoff_ms,
+            backoff_multiplier=args.retry_backoff_multiplier,
+            jitter_factor=args.retry_jitter_factor,
```
waelrabah11:
```suggestion
            max_retries=args.max_retries,
            initial_backoff_ms=args.initial_backoff_ms,
            max_backoff_ms=args.max_backoff_ms,
            backoff_multiplier=args.backoff_multiplier,
            jitter_factor=args.jitter_factor,
```

src/vllm_router/app.py
```diff
@@ -279,7 +293,6 @@ def initialize_all(app: FastAPI, args):
         prefill_model_labels=args.prefill_model_labels,
         decode_model_labels=args.decode_model_labels,
         kv_aware_threshold=args.kv_aware_threshold,
-        max_instance_failover_reroute_attempts=args.max_instance_failover_reroute_attempts,
```
waelrabah11: Why did you remove this ?

src/vllm_router/parsers/parser.py
```diff
@@ -447,11 +461,45 @@ def parse_args():
...
+        "Retry Configuration",
+        "Configure retry behavior with exponential backoff (disabled by default)",
+    )
+    retry_group.add_argument(
+        "--enable-retries",
+        action="store_true",
+        help="Enable automatic retry for transient HTTP failures (408, 429, 500, 502, 503, 504). Disabled by default for fast failover.",
+    )
+    retry_group.add_argument(
+        "--retry-max-retries",
```
waelrabah11:
```suggestion
        "--max-retries",
```

src/vllm_router/parsers/parser.py
```diff
@@ -447,11 +461,45 @@ def parse_args():
...
         type=int,
-        default=0,
-        help="Number of reroute attempts per failed request",
+        default=5,
+        help="Maximum total attempts including initial request (default: 5). Only used when --enable-retries is set.",
+    )
+    retry_group.add_argument(
+        "--retry-initial-backoff-ms",
```
waelrabah11:
```suggestion
        "--initial-backoff-ms",
```

src/vllm_router/parsers/parser.py
```diff
@@ -447,11 +461,45 @@ def parse_args():
...
+        type=int,
+        default=50,
+        help="Initial backoff duration in milliseconds (default: 50)",
+    )
+    retry_group.add_argument(
+        "--retry-max-backoff-ms",
```
waelrabah11:
```suggestion
        "--max-backoff-ms",
```

src/vllm_router/parsers/parser.py
```diff
@@ -447,11 +461,45 @@ def parse_args():
...
+        type=int,
+        default=30000,
+        help="Maximum backoff duration in milliseconds (default: 30000)",
+    )
+    retry_group.add_argument(
+        "--retry-backoff-multiplier",
```
waelrabah11:
```suggestion
        "--backoff-multiplier",
```

src/vllm_router/parsers/parser.py
```diff
@@ -447,11 +461,45 @@ def parse_args():
...
+        type=float,
+        default=1.5,
+        help="Exponential backoff multiplier (default: 1.5)",
+    )
+    retry_group.add_argument(
+        "--retry-jitter-factor",
```
waelrabah11:
```suggestion
        "--jitter-factor",
```

waelrabah11 (commented):
Overall looks good to me. You can consider my comments as nitpicks but I just disagree with using the `retry` prefix. It's a bit redundant and could be avoided with documentation

PR #949, opened 2026-05-11
[Feat] Add per-model custom labels and podLabels to modelSpec

### LINKED ISSUE #934: feature: Additional labels per model deployment via modelSpec
### Describe the feature

Currently, deployments per model via `modelSpec` are deployed with the default standard labels that applies to all model deployments.

This can be used for further observability via collecting pod metadata and also scheduling pods on the desired node types based on internal platform conventions.

We have an internal GPU as a service system that works based on apply a gpu-size label to inference workloads. Our platform then automatically applies nodeaffinity rules, toleration/selectors to the pods to get them scheduled on the matching nodes.

We also use this and a few other labels for observability based on pod metadata to understand how we are using our ML fleet.

### Additional context

I know custom annotations are supported but it would be great to support labels as well.

NasAmin:
Add support for custom `labels` and `podLabels` on a per-model basis via `modelSpec`, similar to the existing `annotations` and `podAnnotations` fields.

`labels` are applied to the Deployment resource metadata, and `podLabels` are applied to the pod template metadata. Neither is added to `selector.matchLabels` since selectors are immutable after creation.

**Use case:** Enables teams to apply custom labels per model for GPU scheduling (e.g., nodeAffinity rules driven by labels) and observability via pod metadata collection.

**Changes:**
- `helm/values.yaml` — Added `labels` and `podLabels` fields to modelSpec
- `helm/values.schema.json` — Added schema definitions for both fields
- `helm/templates/deployment-vllm-multi.yaml` — Render per-model labels on deployment and pod template metadata
- `helm/tests/deployment-vllm-multi_test.yaml` — Added test verifying labels, podLabels, and selector isolation

FIX #934

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -82,6 +82,8 @@ This table documents all available configuration values for the Production Stack
 |-------|------|---------|-------------|
 | `servingEngineSpec.modelSpec[].annotations` | map | `{}` | (Optional) Annotations to add to the deployment, e.g., {model: "opt125m"} |
 | `servingEngineSpec.modelSpec[].podAnnotations` | map | `{}` | (Optional) Annotations to add to the pod, e.g., {model: "opt125m"} |
+| `servingEngineSpec.modelSpec[].labels` | map | `{}` | (Optional) Additional labels to add to the deployment |
+| `servingEngineSpec.modelSpec[].podLabels` | map | `{}` | (Optional) Additional labels to add to the pods |
 | `servingEngineSpec.modelSpec[].name` | string | `""` | The name of the model, e.g., "example-model" |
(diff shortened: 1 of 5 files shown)


ruizhang0101 (commented):
Could you add this to readme as well? Otherwise LGTM.

NasAmin:
> Could you add this to readme as well? Otherwise LGTM.

Thanks, readme updated

ruizhang0101:
@NasAmin Could you fix the pre-commit when you have time?

NasAmin:
> @NasAmin Could you fix the pre-commit when you have time?

@ruizhang0101 Sorry for the late reply. I have pushed a change but not sure if is successful as CI is blocked.

ruizhang0101 (approved):
LGTM

PR #968, opened 2026-06-04
[Bugfix] Propagate serviceAccountName to Ray head and worker pods

ghkdwlgns612:
## Summary

This PR fixes an issue where the serviceAccountName configured in modelSpec was not propagated to the Ray head and worker pods generated by **helm/templates/ray-cluster.yaml**.

As a result, during multi-node Ray deployments, both head and worker pods were created with the default Kubernetes service account instead of the user-specified service account. This caused authentication and authorization failures when accessing cloud resources and other Kubernetes-integrated services.

This change ensures that the serviceAccountName defined in modelSpec is applied consistently to both head and worker pod templates.

## Root Cause

The Ray cluster Helm template did not pass the configured serviceAccountName from modelSpec to the generated pod specifications for the Ray head and worker groups.

Because of this, Kubernetes automatically assigned the default service account to the pods.

## Changes

- Apply the configured service account to Ray head pods
- Apply the configured service account to Ray worker pods

### DIFF
--- a/helm/templates/ray-cluster.yaml
+++ b/helm/templates/ray-cluster.yaml
@@ -26,7 +26,7 @@ spec:
           {{- include "chart.engineLabels" . | nindent 10 }}
       spec:
         terminationGracePeriodSeconds: 0
-        {{- if hasKey $modelSpec "serviceAccountName" }}
+        {{- if $modelSpec.serviceAccountName }}
         serviceAccountName: {{ $modelSpec.serviceAccountName }}
         {{- end }}
         {{- with .Values.servingEngineSpec.securityContext }}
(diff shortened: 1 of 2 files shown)


ruizhang0101 (approved):
LGTM

ruizhang0101:
Hi, could you sigh the commit?

ghkdwlgns612:
Thank you for reviewing the message. I signed it.

ruizhang0101:
It seems like the DCO still didn't pass, could you check it again?

ghkdwlgns612:
I modified it to pass the DCO.

ruizhang0101 (approved):
LGTM

PR #970, opened 2026-06-04
fix(utils): make install-minikube-cluster.sh work on macOS

### LINKED ISSUE #931: Bug: `install-minikube-cluster.sh` fails on macOS (wrong Minikube binary, Linux-only assumptions)
## Environment

- **OS:** macOS (Darwin), Apple Silicon (`arm64`) — e.g. Darwin 24.x  
- **Script:** `utils/install-minikube-cluster.sh`

## What I did

1. Ran: `bash install-minikube-cluster.sh`
2. Script reported Minikube as already installed, then failed when starting the cluster.

## What happened

```text
Minikube already installed.
...
Starting minikube without GPU support...
install-minikube-cluster.sh: line 134: /usr/local/bin/minikube: cannot execute binary file
```

`/usr/local/bin/minikube` was a **Linux ELF** binary (or otherwise non-macOS), so the shell could not execute it on Darwin.

## Why this is confusing

1. **“Already installed” short-circuits install**  
   If `minikube` is on `PATH`, the script never runs the install block, so a **wrong** binary (e.g. Linux `minikube` copied to `/usr/local/bin`) is never replaced.

2. **Install block is Linux-only**  
   When install does run, it only downloads `minikube-linux-amd64`, which is incorrect for macOS (should be e.g. `minikube-darwin-arm64` / `minikube-darwin-amd64`).

3. **Other Linux-only behavior on macOS** (likely follow-on problems even after Minikube is fixed)  
   - `calculate_safe_memory` reads `/proc/meminfo` (not present on macOS) unless `MINIKUBE_MEM` is set.  
   - GPU/Docker path uses `systemctl restart docker` (not applicable on macOS).  
   - BPF section is gated on Linux `/proc` paths (already skipped on Mac, which is fine).

## What I expected

- Either the script **documents** that it targets Linux only, **or** it detects macOS and installs the correct Minikube artifact / points users to `brew install minikube`, and avoids Linux-only memory/GPU/Docker steps.

## Suggested directions (for maintainers)

- **Option A:** Document clearly: “Supported: Linux only; on macOS use …”  
- **Option B:** Add OS detection: on Darwin, skip `calculate_safe_memory` unless `MINIKUBE_MEM` is set, install `minikube-darwin-{arm64,amd64}` when missing, and don’t use `systemctl` for Docker on Mac.

## Workaround that worked for me

- Remove the bad binary: `sudo rm /usr/local/bin/minikube` (after verifying with `file`)  
- Install macOS Minikube: e.g. `brew install minikube`  
- Set `MINIKUBE_MEM` if the script still hits `/proc/meminfo` on macOS until the script is updated.

## What I did

1. Ran: `bash install-minikube-cluster.sh`
2. Script reported Minikube as already installed, then failed when starting the cluster.

## What happened

```text
Minikube already installed.
...
Starting minikube without GPU support...
install-minikube-cluster.sh: line 134: /usr/local/bin/minikube: cannot execute binary file
```

`/usr/local/bin/minikube` was a **Linux ELF** binary (or otherwise non-macOS), so the shell could not execute it on Darwin.

### Expected behavior

_No response_

### Additional context

_No response_

HumphreySun98:
Fixes #931.

### Summary

The script assumed Linux throughout and failed on macOS in three distinct ways. All are fixed in this PR without changing Linux behavior.

### Fixes

1. **Minikube binary selection.** Previously downloaded \`minikube-linux-amd64\` unconditionally even on Darwin. Also short-circuited on \"minikube already on PATH\" without checking the binary is actually executable — a Linux ELF copied to \`/usr/local/bin/minikube\` would never get replaced (matches the issue's reproduction: \`/usr/local/bin/minikube: cannot execute binary file\`). Now detects OS (\`uname -s\`) and arch (\`uname -m\`), downloads the matching \`minikube-{linux,darwin}-{amd64,arm64}\` release, and requires the existing binary to actually \`minikube version\` cleanly before treating it as installed.

2. **\`calculate_safe_memory\`.** Read \`/proc/meminfo\`, which doesn't exist on macOS. Now uses \`sysctl -n hw.memsize\` on Darwin (with a ~60% available estimate, since Docker Desktop's VM budget has no clean equivalent to Linux's \`MemAvailable\`) and keeps the existing \`/proc/meminfo\` path on Linux. cgroup-v2 detection stays Linux-only.

3. **\`systemctl restart docker\` / \`sudo sysctl fs.protected_regular=0\`.** Both Linux-only — \`systemctl\` isn't on macOS, and \`fs.protected_regular\` is a Linux sysctl. Now guarded on \`HOST_OS == linux\`, with a hint on macOS to restart Docker Desktop manually.

The BPF section was already guarded on \`/proc/sys/net/core/bpf_jit_harden\` existence, so no change needed there.

### Test plan

- \`bash -n utils/install-minikube-cluster.sh\` (syntax check) — passes.
- Reading the diff: every Linux-only path is now behind an \`HOST_OS == linux\` (or pre-existing \`/proc\` existence) gate, so Linux behavior is byte-for-byte identical when those guards evaluate the same way.
- On macOS, the previously-failing path now downloads \`minikube-darwin-arm64\` (or \`-amd64\` on Intel), skips \`systemctl\` and \`sysctl fs.protected_regular\`, and uses \`sysctl -n hw.memsize\` for memory sizing.

### Out of scope

Docker Desktop restart automation on macOS (would need \`osascript\` + user permission). Left as a printed instruction. Same for the GPU Operator step — NVIDIA GPU passthrough on Apple Silicon Docker Desktop isn't supported by NVIDIA Container Toolkit, so users on macOS will hit the \`No NVIDIA GPU detected\` branch anyway.


ruizhang0101 (dismissed):
LGTM

HumphreySun98:
Thanks for the review! Addressed both points:

- **`curl -fLO`** — applied (1b1c450). Agreed; without `-f`, curl saves the error page and exits 0, so the script would try to install HTML as the binary.
- **`install-kubectl.sh` hardcodes `linux/amd64`** — good catch, and a real gap for a fresh macOS host (it does early-exit when `kubectl` is already on `PATH`, e.g. via brew, so it isn't always hit). Since it's a separate script and outside the scope of #931 (which is specifically `install-minikube-cluster.sh`), I'll fix it in a small follow-up PR that gives `install-kubectl.sh` the same `uname`-based OS/arch detection — happy to link it here once it's up.

Also updated the branch to latest `main` (it had gone `BEHIND`). @ruizhang0101 — rebased + review feedback addressed, ready when you are. Thanks!

HumphreySun98:
Follow-up for the `install-kubectl.sh` OS/arch point is up: #976.

HumphreySun98:
Thanks for merging #976! I noticed #970 was closed but #931 (the `install-minikube-cluster.sh` macOS support) is still open — should I reopen/rebase this one, or did the minikube changes land elsewhere? Happy to resubmit a clean rebase if it's still useful.

ruizhang0101 (approved):
LGTM

PR #976, opened 2026-06-16
fix(utils): make install-kubectl.sh work on macOS

HumphreySun98:
## Summary

Follow-up to the review on #970 (and same macOS root cause as #931): `utils/install-kubectl.sh` always downloaded the `linux/amd64` kubectl, so on macOS it installs an unexecutable Linux ELF and later `kubectl` calls fail with an exec-format error. #970 fixed `install-minikube-cluster.sh`, which calls this script — so without this change the end-to-end flow still breaks on macOS at the kubectl step.

## Changes

- Detect OS (`uname -s` → linux/darwin) and arch (`uname -m` → amd64/arm64) and download the matching `bin/{linux,darwin}/{amd64,arm64}/kubectl`, mirroring the detection added to `install-minikube-cluster.sh` in #970.
- Treat an existing kubectl as installed only if it actually runs (`kubectl version --client`), so a stale Linux binary left on `PATH` on macOS is replaced instead of skipped (this was the exact `#931` failure mode).
- Use `curl -f` so an HTTP error aborts the script (via `set -e`) instead of saving an error page as the binary.

Linux behavior is unchanged: `HOST_OS`/`HOST_ARCH` resolve to `linux`/`amd64` on a normal x86_64 Linux host, giving the same URL as before.

## Test plan

- `bash -n utils/install-kubectl.sh` — passes (shellcheck runs in CI).
- On Linux x86_64: resolves to `bin/linux/amd64/kubectl` (identical to the previous hardcoded URL).
- On macOS arm64: resolves to `bin/darwin/arm64/kubectl`; a previously-installed Linux binary on `PATH` no longer short-circuits the install.

AI assistance was used; I have reviewed, run, and can defend every change.

### DIFF
--- a/utils/install-kubectl.sh
+++ b/utils/install-kubectl.sh
@@ -5,11 +5,30 @@ set -e
 KUBECTL_DIR="$HOME/.local/bin"
 KUBECTL_PATH="$KUBECTL_DIR/kubectl"
 
+# Detect host OS and architecture so we download the matching kubectl build.
+# Without this the script always pulled the linux/amd64 binary, which on macOS
+# installs an unexecutable Linux ELF (exec format error on later kubectl calls).
+OS_KERNEL="$(uname -s)"
+case "$OS_KERNEL" in
+  Linux)  HOST_OS=linux ;;
+  Darwin) HOST_OS=darwin ;;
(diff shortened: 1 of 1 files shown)


ruizhang0101 (approved):
LGTM

PR #980, opened 2026-06-21
feat(helm/raycluster): pass through annotations and podAnnotations (#962)

### LINKED ISSUE #962: feature: Support adding modelSpec pod annotations to ray-cluster spec.
### Describe the feature

Add the block in the raySpec for passing through annotations. It appears the values exist, they just only apply to multi vllm deployment, not kuberay.

Adding multus and other things to the pods.

### Additional context

It seems like it exists but wasn't fully finished. Curious if maybe there was a reason.

Anai-Guo:
## Summary

`modelSpec.annotations` and `modelSpec.podAnnotations` already exist and are honored for the standard (non-Ray) vLLM `Deployment` (`deployment-vllm-multi.yaml`), but the **RayCluster** template never consumed them. Users running engines through KubeRay (`raySpec.enabled: true`) therefore had no way to attach pod-level annotations — e.g. Multus `k8s.v1.cni.cncf.io/networks` — or object-level annotations to their head/worker pods. As #962 notes, the values appear to exist but were simply never wired into the kuberay path.

## Changes

`helm/templates/ray-cluster.yaml`, mirroring the existing `Deployment` pattern:

- `modelSpec.annotations` → `RayCluster` object `metadata.annotations`
- `modelSpec.podAnnotations` → head pod template `metadata.annotations`
- `modelSpec.podAnnotations` → worker pod template `metadata.annotations`

No new values are introduced — this only completes the existing passthrough for the KubeRay path. Both blocks are guarded with `{{- with ... }}`, so output is unchanged when the values are empty (the default).

## Test

Adds a `helm-unittest` case to `helm/tests/ray-cluster_test.yaml` asserting that both `annotations` (object) and `podAnnotations` (head + worker) render on the RayCluster.

Closes #962

### DIFF
--- a/helm/templates/ray-cluster.yaml
+++ b/helm/templates/ray-cluster.yaml
@@ -9,6 +9,10 @@ kind: RayCluster
 metadata:
   name: "{{ .Release.Name }}-{{$modelSpec.name}}-raycluster"
   namespace: {{ .Release.Namespace }}
+  {{- with $modelSpec.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
   labels:
     model: {{ $modelSpec.name }}
     helm-release-name: {{ .Release.Name }}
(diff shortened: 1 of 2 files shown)


ruizhang0101 (approved):
LGTM

PR #860, opened 2026-02-27
feat(helm) add monitoring conf as a sub chart

enneitex:
Contributes to https://github.com/vllm-project/production-stack/issues/840

The goal of this PR is to be more flexible about the monitring part of this chart. Currently there is no way to use an existing prometheus stack.
By adding a few resources in the chart and prometheus as a sub-chart, it allows user to fully customize their installation, from only a `serviceMonitor` to a whole prometheus stack

**Technical changes:**

- Add `serviceMonitor` directly to the helm chart.
- Add grafana dashboards to the helm chart.
- Add kube-prometheus-stack and prometheus-adapter as sub-charts.
- Update the documentation that mentioned the old installation method.
   

**Step to validate the PR:**

1) Install the helm chart with the following values (+ your values with modelSpec etc...) 

```yaml
servingEngineSpec:
  serviceMonitor:
    enabled: true
routerSpec:
  serviceMonitor:
    enabled: true
kube-prometheus-stack:
  enabled: true
prometheus-adapter:
  enabled: true
```

2) get the admin password for grafana, stored in a secret `<release-name>-grafana`

3) Port foward grafana UI and login: `kubectl port-forward svc/<release-name>-grafana 8080:80`. Two vLLM dashboards should be present, with data if there are request made to the models

```bash
helm install llmstack . -f observability.yaml
kubectl get secret llmstack-grafana -oyaml
kubectl port-forward service/llmstack-grafana 8080:80
```

- [x] Make sure the code changes pass the [pre-commit](https://github.com/vllm-project/production-stack/blob/main/CONTRIBUTING.md) checks.
- [x] Sign-off your commit by using <code>-s</code> when doing <code>git commit</code>
- [x] Try to classify PRs for easy understanding of the type of changes, such as `[Bugfix]`, `[Feat]`, and `[CI]`.

<details>

<summary><b> Detailed Checklist (Click to Expand) </b></summary>

<p>Thank you for your contribution to production-stack! Before submitting the pull request, please ensure the PR meets the following criteria. This helps us maintain the code quality and improve the efficiency of the review process.</p>

<h3>PR Title and Classification</h3>
<p>Please try to classify PRs for easy understanding of the type of changes. The PR title is prefixed appropriately to indicate the type of change. Please use one of the following:</p>
<ul>
    <li><code>[Bugfix]</code> for bug fixes.</li>
    <li><code>[CI/Build]</code> for build or continuous integration improvements.</li>
[...]

PR #982, opened 2026-06-24
docs(helm): restore LMCache dashboard explanation in Observability section

samhoss93:
## What

Adds back a short explanation of the LMCache dashboard's six fields to `helm/README.md`'s Observability section.

## Why

The old `observability/README.md` documented this under "Import LMCache Dashboard," and is now deprecated with a banner pointing readers to `helm/README.md#observability` for the same content. During the monitoring refactor in #860, the dashboard itself (`helm/dashboards/lmcache-dashboard.json`) moved into the chart and is now auto-provisioned via `helm/templates/dashboards.yaml` — a real improvement, no more manual `kubectl apply` step. But the explanation of what the dashboard actually shows didn't make the move over, so the deprecation banner currently points to a section that doesn't have it.

This PR only adds the explanatory text (adapted to note it's auto-provisioned now). It intentionally drops the old manual `kubectl apply`/restart commands from the original doc, since those are obsolete now that the chart handles provisioning.

## Scope

Single doc change, no code/behavior change.


ruizhang0101 (approved):
LGTM

PR #985, opened 2026-07-01
fix(operator): tolerate missing KEDA when cleaning up ScaledObject (#981)

### LINKED ISSUE #981: bug: The VLLMRuntime controller fails on every reconciliation when running on a cluster that does not have KEDA installed
### Describe the bug

The VLLMRuntime controller fails on every reconciliation when running on a cluster that does not have KEDA installed, even if no VLLMRuntime resource enables autoscaling. This makes the operator unusable on clusters without KEDA.

#### Environment 
- vLLM Production Stack Operator
- Cluster without KEDA (no `keda.sh` API group registered)
#### Steps to Reproduce
1. Install the operator on a cluster **without** KEDA
2. Create a VLLMRuntime resource **without** `autoscalingConfig` (or with `autoscalingConfig.enabled: false`)
3. Observe the controller logs

### Expected behavior

The VLLMRuntime reconciles successfully. Since autoscaling is not requested, the operator should not require KEDA to be present.

### Additional context

Root cause:

In `operator/internal/controller/vllmruntime_controller.go:360-370`, when `autoscalingConfig` is nil or disabled, the `else` branch attempts to delete a ScaledObject as a cleanup measure:

https://github.com/vllm-project/production-stack/blob/9b78da7dd9df3652bea4cddaa31d9087497cebc8/operator/internal/controller/vllmruntime_controller.go#L360-L370

The guard `!errors.IsNotFound(err)` only catches the case where the ScaledObject **resource** does not exist (i.e., the `keda.sh` API group is registered, but no ScaledObject with that name is found). It does **not** catch the case where the `keda.sh` API group itself is not registered on the cluster. In that scenario, the API server returns a different error (e.g., `"the server could not find the requested resource"`) that does not satisfy `errors.IsNotFound()`, causing the reconciliation to fail.

Anai-Guo:
## Summary

Fixes #981.

The `VLLMRuntime` controller failed on **every** reconciliation on clusters without KEDA installed, even when no `VLLMRuntime` requested autoscaling — making the operator unusable there.

### Root cause

In the autoscaling-disabled branch of `Reconcile`, the controller best-effort deletes a possibly-stale `keda.sh/v1alpha1` `ScaledObject`:

```go
if err := r.Delete(ctx, scaledObject); err != nil && !errors.IsNotFound(err) {
    return ctrl.Result{}, err
}
```

When KEDA is not installed, the `keda.sh` API group is not registered, so the RESTMapper returns a `*meta.NoKindMatchError` — **not** a `NotFound`. The existing guard only tolerates `errors.IsNotFound(err)`, so the reconcile errored out and requeued forever.

### Fix

Also tolerate `meta.IsNoMatchError(err)` on this cleanup path. If the CRD is absent there is nothing to delete, so the reconcile should succeed. The autoscaling-**enabled** path (`reconcileScaledObject`) is unchanged — enabling autoscaling legitimately requires KEDA.

Two-line change plus an explanatory comment; no behavior change on clusters that do have KEDA.

## Testing

- Verified the guard now short-circuits on `NoKindMatchError` (returned by controller-runtime's client when the `keda.sh/v1alpha1` group/kind is unregistered), which is precisely the KEDA-absent case.


aeon-x (approved):

aeon-x:
LGTM

PR #987, opened 2026-07-01
[Bugfix][Helm] skip lmcache dashboard when cache server is disabled

bitnik:

When `grafanaDashboards.enabled` is true, the chart renders a `ConfigMap` for every JSON file in `dashboards/*.json`, including `lmcache-dashboard.json`. This happened regardless of whether the cache server was actually deployed, so clusters running with `cacheserverSpec.enabled: false` ended up with an LMCache Grafana dashboard for a component that isn't running  and it showes panels with no-data.

This PR makes the dashboard rendering conditional: `lmcache-dashboard.json` is now skipped when `cacheserverSpec.enabled` is `false`. All other dashboards continue to render as before.


ruizhang0101 (approved):
LGTM

PR #988, opened 2026-07-01
feat(helm): support externally-managed PVC via pvcExistingClaimName (#938)

### LINKED ISSUE #938: feature: Existing Claim for PVC
### Describe the feature

I would like to manage the PVC separate from the chart and use existingClaims to fix up.

I use a GitOps system for a controlled environment. The PVC policy for the usage of helm charts is to have them managed outside of chart and then brought in. 

### Additional context

_No response_

Anai-Guo:
## Summary

Fixes #938.

Adds an optional per-model `pvcExistingClaimName` to `servingEngineSpec.modelSpec[]`, so operators can mount a `PersistentVolumeClaim` that is managed **outside** the chart instead of one the chart creates. This is useful for GitOps environments where PVC policy requires claims to be provisioned separately and brought in via `existingClaim`.

## Behavior

- **`pvcExistingClaimName` set** → the chart skips rendering the model PVC (`pvc.yaml`) and the serving-engine deployment mounts the named claim directly.
- **`pvcExistingClaimName` unset (default)** → unchanged: the chart creates and mounts `<release>-<model>-storage-claim`.

## Changes

- `helm/templates/pvc.yaml` — guard PVC creation with `(empty $modelSpec.pvcExistingClaimName)`.
- `helm/templates/deployment-vllm-multi.yaml` — use `pvcExistingClaimName` for the volume `claimName` when set, falling back to the generated name.
- `helm/values.yaml` — document the option as a commented example. Kept as a plain comment (not a `# --` helm-docs key) so the generated `values.schema.json` is unaffected; since `modelSpec` items do not set `additionalProperties: false`, the new key still validates.
- `helm/tests/pvc_test.yaml` — add a case asserting no PVC document is rendered when `pvcExistingClaimName` is set.

Scope is limited to the per-model storage claim referenced by the serving-engine deployment; the separate `sharedPvcStorage` path is unchanged.

### DIFF
--- a/helm/templates/deployment-vllm-multi.yaml
+++ b/helm/templates/deployment-vllm-multi.yaml
@@ -471,7 +471,7 @@ spec:
             {{- toYaml $modelSpec.pvcStorage.emptyDir | nindent 12 }}
         {{- else if kindIs "string" $modelSpec.pvcStorage }}
           persistentVolumeClaim:
-            claimName: "{{ .Release.Name }}-{{$modelSpec.name}}-storage-claim"
+            claimName: {{ $modelSpec.pvcExistingClaimName | default (printf "%s-%s-storage-claim" .Release.Name $modelSpec.name) | quote }}
         {{- end }}
         {{- end }}
         {{- if $.Values.sharedPvcStorage.enabled }}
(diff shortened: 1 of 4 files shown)


ruizhang0101 (approved):
LGTM

PR #992, opened 2026-07-07
[Feat][Helm] Support customizable resources and init containers for RayCluster

TheCodingSheikh:
Extends the Ray cluster template so users can now:

- Provide a raw Kubernetes resources block on both the Ray head (raySpec.headNode.resources) and the Ray workers (modelSpec.resources). When set, the raw block takes priority over the simplified request*/limit* fields, mirroring the existing behavior in deployment-vllm-multi.yaml.
- Attach an init container to the Ray head pod via raySpec.headNode.initContainer and to each Ray worker pod via modelSpec.initContainer, using the same block shape as the regular vLLM deployment (name, image, command, args, env, resources, mountPvcStorage, extraVolumeMounts).

The values.yaml documentation, values.schema.json and helm/README.md are updated so contributors can discover the new fields, and helm/tests/ray-cluster_test.yaml is extended with helm-unittest cases that cover the new head/worker init containers, the raw resources overrides, and the fall-through to the simplified request* fields.

Bumps helm/Chart.yaml to 0.1.12 per the chart's version-bump instructions.

The change is backwards compatible: existing raySpec configurations without initContainer or resources render identically to before.

### DIFF
--- a/helm/Chart.yaml
+++ b/helm/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.1.11
+version: 0.1.12
 
 maintainers:
   - name: apostac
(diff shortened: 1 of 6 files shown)


ruizhang0101 (approved):
LGTM

PR #993, opened 2026-07-11
fix(operator): use per-router names for VLLMRouter pod-viewer RBAC

### LINKED ISSUE #900: bug: use multiple crd vllmrouter in the same namespace
### Describe the bug

hi

Pod's names pod-viewer-role and pod-viewer-binding are hardcoded so we cannot have more than 1 vllmrouter in the same namespace.

### To Reproduce

Try to deploy 2 crd vllmrouter in the same namespace for 2 differents models.

### Expected behavior

_No response_

### Additional context

_No response_

Anai-Guo:
## What

`VLLMRouter` reconciliation creates a `Role` and `RoleBinding` (granting `get/list/watch` on pods) using **fixed names** `pod-viewer-role` / `pod-viewer-binding`, both owner-referenced to the creating router. This makes it impossible to run more than one `VLLMRouter` in the same namespace (#900):

- The **second** router finds the existing `RoleBinding` (creation is skipped) whose only subject is the **first** router's ServiceAccount — so the second router's SA is never bound and its pods cannot list pods.
- Because the shared `Role`/`RoleBinding` are owned by the router that first created them, **deleting that router garbage-collects them**, breaking the surviving router.

## Change

Derive both RBAC object names from the router name via two small helpers:

```go
func podViewerRoleName(router *servingv1alpha1.VLLMRouter) string {
    return fmt.Sprintf("%s-pod-viewer-role", router.Name)
}
func podViewerBindingName(router *servingv1alpha1.VLLMRouter) string {
    return fmt.Sprintf("%s-pod-viewer-binding", router.Name)
}
```

Each router now gets its own `Role` and `RoleBinding` bound to its own ServiceAccount, so multiple routers coexist in one namespace and lifecycle events on one no longer affect the others. The `RoleRef` in the binding is updated to point at the per-router role name.

No behavior change for single-router namespaces (name is just prefixed). `fmt` is already imported; no test references the old fixed names.

Fixes #900

### DIFF
--- a/operator/internal/controller/vllmrouter_controller.go
+++ b/operator/internal/controller/vllmrouter_controller.go
@@ -96,7 +96,7 @@ func (r *VLLMRouterReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 
 	// Create Role if it doesn't exist
 	role := &rbacv1.Role{}
-	err = r.Get(ctx, types.NamespacedName{Name: "pod-viewer-role", Namespace: router.Namespace}, role)
+	err = r.Get(ctx, types.NamespacedName{Name: podViewerRoleName(router), Namespace: router.Namespace}, role)
 	if err != nil && errors.IsNotFound(err) {
 		role = r.roleForVLLMRouter(router)
 		log.Info("Creating a new Role", "Role.Namespace", role.Namespace, "Role.Name", role.Name)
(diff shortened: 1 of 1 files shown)


ruizhang0101 (approved):
LGTM

PR #994, opened 2026-07-12
feat(operator): support shmSize on VLLMRuntime for tensor parallelism (#899)

### LINKED ISSUE #899: bug: CRD VllmRuntime : increase shm size
### Describe the bug

Hi

Currently we cannot increase the shm size like the helm chart for tensor parallelism

### To Reproduce

Find shm size option in crd api

### Expected behavior

Add shmSize: 24g

### Additional context

_No response_

Anai-Guo:
## What

Adds a `shmSize` field to the `VLLMRuntime` CRD (`deploymentConfig.shmSize`). When set, the operator mounts an `emptyDir` with `medium: Memory` at `/dev/shm`, sized to the given quantity.

Fixes #899.

## Why

Tensor-parallel vLLM communicates between ranks over shared memory. In containers `/dev/shm` defaults to a small size (typically 64Mi), which is too small for TP and leads to crashes / `Bus error`. The Helm chart can be worked around with extra volumes, but the `VLLMRuntime` operator had no equivalent knob — exactly what the issue asks for (`shmSize: 24g`).

## What changed

- `api/v1alpha1/vllmruntime_types.go` — new optional `ShmSize string` on `DeploymentConfig`.
- `internal/controller/vllmruntime_controller.go` — when `ShmSize != ""`, append a `dshm` `emptyDir{medium: Memory}` volume + a `/dev/shm` mount on the vLLM container. The size is parsed with `resource.ParseQuantity`; an unparseable value still yields a Memory-backed `/dev/shm` (no size limit) rather than panicking the reconcile loop.
- `config/crd/bases/...vllmruntimes.yaml` — regenerated CRD property.

## Behavior

- Field unset → no change (backwards compatible; `dshm` volume is only added when requested).
- `shmSize: "24Gi"` → `/dev/shm` backed by a 24Gi Memory `emptyDir`.

Example:

```yaml
spec:
  deploymentConfig:
    shmSize: "24Gi"
```

### DIFF
--- a/operator/internal/controller/vllmruntime_controller.go
+++ b/operator/internal/controller/vllmruntime_controller.go
@@ -738,6 +738,23 @@ func (r *VLLMRuntimeReconciler) deploymentForVLLMRuntime(
 		})
 	}
 
+	// Mount an emptyDir (medium=Memory) at /dev/shm when a size is requested.
+	// Tensor parallelism communicates over shared memory, and the container
+	// default /dev/shm is usually too small for it.
+	if vllmRuntime.Spec.DeploymentConfig.ShmSize != "" {
+		shmSource := corev1.VolumeSource{
+			EmptyDir: &corev1.EmptyDirVolumeSource{Medium: corev1.StorageMediumMemory},
+		}
+		if q, err := resource.ParseQuantity(vllmRuntime.Spec.DeploymentConfig.ShmSize); err == nil {
+			shmSource.EmptyDir.SizeLimit = &q
+		}
+		volumes = append(volumes, corev1.Volume{Name: "dshm", VolumeSource: shmSource})
+		volumeMounts = append(volumeMounts, corev1.VolumeMount{
+			Name:      "dshm",
+			MountPath: "/dev/shm",
+		})
+	}
+
 	var affinity *corev1.Affinity
 
 	if vllmRuntime.Spec.DeploymentConfig.NodeSelectorTerms != nil {
--- a/operator/internal/controller/vllmruntime_controller.go
+++ b/operator/internal/controller/vllmruntime_controller.go
@@ -1111,17 +1111,10 @@ func (r *VLLMRuntimeReconciler) deploymentNeedsUpdate(
 		}
 	}
 
-	// Detect drift in the /dev/shm volume (driven by shmSize). Without this,
-	// toggling shmSize on an existing VLLMRuntime is seen as "no change" and
-	// the running Deployment never picks up (or drops) the mount.
-	//
-	// We compare the "dshm" volume/mount specifically rather than the whole
-	// Volumes slice: the API server defaults fields on some volume sources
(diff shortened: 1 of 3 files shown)


ruizhang0101:
Could you resolve the suggestions from gemini?

Anai-Guo:
Thanks @ruizhang0101 — addressed the gemini review in cd3c1ee:

**High — shmSize changes not triggering a Deployment update:** `deploymentNeedsUpdate` now detects drift in the `/dev/shm` volume and mount, so editing `shmSize` on an existing `VLLMRuntime` actually rolls the Deployment. I compared the `dshm` volume/mount **specifically** rather than the whole `Volumes` slice, because a blanket `reflect.DeepEqual` over all volumes would falsely mismatch on API-server-defaulted fields (e.g. `ConfigMap.DefaultMode` on the chat-template volume), causing a perpetual reconcile loop. The `dshm` emptyDir/mount have no such server-side defaulting, so comparing them directly is safe.

**Medium — silent failure on invalid quantity:** an unparseable `shmSize` is no longer swallowed; it's logged as an error instead of silently mounting an unbounded `/dev/shm`.

**Medium — `*resource.Quantity` vs `string`:** I kept `ShmSize` as a `string` for consistency with the existing resource fields — `cpu`, `gpu`, `memory` in `ResourceRequirements` are all plain strings, so introducing the only `*resource.Quantity`-typed field in the API would be inconsistent. The invalid-value concern is now surfaced via the warning log above. Happy to switch to `*resource.Quantity` (and regenerate the CRD) if you'd prefer the API-level validation despite the inconsistency — just let me know.

operator/internal/controller/vllmruntime_controller.go
```diff
@@ -1087,9 +1111,57 @@ func (r *VLLMRuntimeReconciler) deploymentNeedsUpdate(
...
+	// the running Deployment never picks up (or drops) the mount.
+	//
+	// We compare the "dshm" volume/mount specifically rather than the whole
+	// Volumes slice: the API server defaults fields on some volume sources
+	// (e.g. ConfigMap DefaultMode, used by the chat-template volume) that the
+	// freshly generated expected spec does not carry, so a blanket
+	// reflect.DeepEqual would report a permanent mismatch and cause an endless
+	// reconcile loop. The dshm emptyDir/mount have no such server-side
+	// defaulting, so comparing them directly is safe.
+	if !reflect.DeepEqual(
```
ruizhang0101: Could you make the docstring concise?

ruizhang0101 (changes requested):
Otherwise LGTM

Anai-Guo:
Thanks @ruizhang0101 — condensed the drift-detection comment to 4 lines (kept the rationale for comparing the `dshm` volume specifically to avoid a reconcile loop from server-side defaulting). PTAL.

ruizhang0101 (approved):
LGTM

PR #996, opened 2026-07-14
[Feat] Add priority routing 

punitvara:
Adds a new proposal at `proposals/priority-routing.md` to define priority-aware routing for the vLLM Production Stack router. The document specifies how request priority is extracted, used for load-aware engine selection, and forwarded for in-engine scheduling.

- **What this PR adds**
  - New proposal file: `proposals/priority-routing.md`
  - Full RFC-style structure: Summary, Motivation, Goals/Non-Goals, Proposal, Drawbacks, Alternatives, Timeline, References

- **Key design decisions captured**
  - Introduces `routing-logic=priority`
  - Defines priority resolution order: header → body field → default
  - Selects **Option A (binary threshold)** as initial strategy for load-ranked routing
  - Specifies fallback to round-robin when load metrics are unavailable
  - Preserves vLLM pass-through semantics (`priority` field injection)

- **Interface/config details documented**
  - Proposed router flags:
    - `--priority-header`
    - `--priority-field`
    - `--priority-default`
    - `--priority-threshold`

Example excerpt from the proposal:

```text
Priority source resolution:
1) --priority-header (default: x-request-priority)
2) --priority-field  (default: priority)
3) --priority-default (default: 0)

Selection rule (Option A):
if priority <= priority-threshold -> least-loaded engine
else -> round-robin over remaining engines
```

Signed-off-by: Punit Vara <punitvara@gmail.com>


proposals/priority-routing.md
```diff
@@ -0,0 +1,95 @@
...
+- `src/vllm_router/parsers/parser.py` — new CLI arguments and `--routing-logic` choice.
+- `src/vllm_router/app.py` and `src/vllm_router/dynamic_config.py` — thread new kwargs through router initialization and dynamic reconfiguration.
+- `src/vllm_router/services/request_service/request.py` — route via `PriorityRouter` and inject the priority field into the forwarded body.
+- **Interface Changes:** New CLI arguments:
+| Argument | Default | Description |
+| -------- | ------- | ----------- |
+| `--routing-logic=priority` | — | Enable priority routing |
+| `--priority-header` | `x-request-priority` | Header carrying the request priority |
+| `--priority-field` | `priority` | Body field carrying the request priority |
+| `--priority-default` | `0` | Priority used when none is provided |
```
ruizhang0101, line 94: Does default 0 mean all the requests have the highest priority? Is this intentional?

proposals/priority-routing.md
```diff
@@ -0,0 +1,95 @@
...
+                                                │
+                                                │ inject “priority”: P into body
+                                                ▼
+                                    ┌────────────────┐
+                                    │ vLLM engine    │ priority scheduler
+                                    │ (preempts)     │ serves P first
+                                    └────────────────┘
+```
+1. Client sends a request with a priority (header or body).
+2. Router extracts the priority and ranks candidate engines by live load (`num_running_requests + num_queuing_requests` from `EngineStats`).
```
ruizhang0101: The enginestats might have delay since it is scraped based. Would be better to use the router side inflight requests metrics.

proposals/priority-routing.md
```diff
@@ -0,0 +1,95 @@
...
+```text
+┌──────────┐     priority=P     ┌───────────────────────────────┐
+│  Client  │───────────────────▶│ Router (routing-logic=priority)│
+└──────────┘                    └───────────────┬───────────────┘
+                                                │ rank engines by load
+                                                │ (num_running + num_queuing)
+                high priority ──────────────────┼────────▶ least-loaded engine
+                normal priority ────────────────┼────────▶ round-robin remainder
+                                                │
+                                                │ inject “priority”: P into body
```
ruizhang0101, line 61: vllm might not be able to take priority if the scheduling is not priority, consider adding a safe fallback and a warning?

proposals/priority-routing.md
```diff
@@ -0,0 +1,95 @@
...
+                                    │ (preempts)     │ serves P first
+                                    └────────────────┘
+```
+1. Client sends a request with a priority (header or body).
+2. Router extracts the priority and ranks candidate engines by live load (`num_running_requests + num_queuing_requests` from `EngineStats`).
+3. The request is mapped to an engine according to the selection strategy below. If no load stats are available, all requests fall back to round-robin.
+4. Router injects `priority` into the forwarded body so vLLM’s priority scheduler can preempt within the engine.
+5. Response streams back normally.
+**Engine selection strategy.** Two strategies were considered for mapping a priority value onto the load-ranked engine list:
+- **Option A — binary threshold (proposed default).** Requests whose priority is at least as important as a configurable threshold (`priority <= --priority-threshold`, default equal to `--priority-default`) are routed to the least-loaded engine; all others round-robin across the remaining engines. Simple to reason about, deterministic, and easy to unit-test. Selected as the default.
```
ruizhang0101: If all the requests are low priority, the least loaded engine will never be used. Consider roundrobin across all the engines instead of remaining engines.

ruizhang0101 (commented):
Thanks for the proposal, I think it is solid overall with some minor suggestions.

proposals/priority-routing.md
punitvara: Good catch — that was unintentional. Fixed by tightening the comparison to a strict `priority < threshold`. Now default-priority (untagged) traffic never auto-qualifies as high-priority — only requests explicitly marked below the threshold do. `--priority-threshold` still defaults to `--priority-default` so the mental model stays simple, but the boundary is now exclusive.

proposals/priority-routing.md
punitvara: Agreed. updated for the same.

proposals/priority-routing.md
punitvara: Added a fallback + warning. if the engine wasn't started with `--scheduling-policy priority`, vLLM just ignores the injected `priority` field with no adverse effect (safe no-op), and the router logs a one-time startup warning reminding operators to verify their engines are configured for priority scheduling when `routing-logic=priority` is on. happy to discuss if we want a stronger guard

proposals/priority-routing.md
punitvara: done.

punitvara:
Should I raise separate PR draft for code ? Or should I keep commit with this PR ?

punitvara:
@ruizhang0101 I updated Proposal and also added code. Can you please review it ?

punitvara:
Do you want me to update anything @ruizhang0101  ?

ruizhang0101 (dismissed):
Could you add some user-facing doc/tutorial? Otherwise LGTM.

punitvara:
ok done @ruizhang0101 Do we need to get any other approval or is it good to merge ?

ruizhang0101 (dismissed):
LGTM

punitvara:
Can you restart worfklows @ruizhang0101 ? Not sure why they are getting timed out. I have rebased and had force push, auto merge was automatically disabled.

punitvara:
@ruizhang0101 Can I focus on any other issue if nothing else is required in this one ?

punitvara:
I think there is some issue with checks. This wont auto merge @ruizhang0101 Can you check that part ?

ruizhang0101 (approved):
LGTM

PR #998, opened 2026-07-16
[Bugfix](router): strip Server header to avoid duplicate headers on response

mashun4ek:
Vllm backend uvicorn sends `server: uvicorn` header. The router copies that into resp_headers, and then the router's own server (uvicorn) appends its Server header when sending the response, so the client sees the Server header twice. Some packages (such as `aiohttp >=3.14`) reject responses with duplicate Server headers, hence the error `Duplicate 'Server' header found.` Let's strip server header from response. 

Let me know if you prefer to add `"server_header": False,` here instead https://github.com/vllm-project/production-stack/blob/90ddf82b97ec9b09098de9976dc8dcb75c6be363/src/vllm_router/app.py#L397-L402


ruizhang0101 (approved):
LGTM

mashun4ek:
looks like can't merge before some workflows will be approved, feel free to merge anytime

PR #999, opened 2026-07-17
[Router][Bugfix] reset K8s watcher on stale resource version errors

loicrouillermonay:
Getting `504 Timeout: Too large resource version` looping in the router logs. vLLM pods were healthy, but the K8s watcher was stuck.

The bug: both `K8sPodIPServiceDiscovery` and `K8sServiceNameServiceDiscovery` create one `kubernetes.watch.Watch` instance and never replace it. The Python client stores the last `resourceVersion` it saw and feeds it back into every reconnect. Once the API server's watch cache moves past that version, the server throws the `504` (or `410 Gone`) and the router catches it, sleeps 0.5s then calls `stream()` again with the same stale bookmark forever.

The Kubernetes Python client docs call this out explicitly in `Watch.stream()` [here](https://github.com/kubernetes-client/python/blob/master/kubernetes/base/watch/watch.py#L154):

> "Note that watching an API resource can expire... if that last result is too old as well, an `ApiException` exception will be thrown... In that case you have to recover yourself, probably by listing the API resource to obtain the latest state..."

So the fix is to reset the watcher when we hit a stale resource version. A new `Watch()` starts with `resource_version=None`, which forces a fresh `LIST` on the next attempt. Then we grab the current bookmark and resume. I only do this for stale-version errors.

Here is below an anonymized snippet so you can see the watcher logs `504` with a stale resource version over and over while the router reports 0 serving engine(s).

```bash
2026-07-17T16:25:53.730Z INFO:     Application startup complete.
2026-07-17T16:25:53.730Z INFO:     Uvicorn running on http://0.0.0.0:8000
2026-07-17T16:25:56.694Z ERROR: K8s watcher error: (504)
Reason: Timeout: Timeout: Too large resource version: <newer>, current: <current>
 (service_discovery.py:721)
 
2026-07-17T16:26:00.208Z ERROR: K8s watcher error: (504)
Reason: Timeout: Timeout: Too large resource version: <newer>, current: <current>
 (service_discovery.py:721)
 
2026-07-17T16:26:03.722Z ERROR: K8s watcher error: (504)
Reason: Timeout: Timeout: Too large resource version: <newer>, current: <current>
 (service_discovery.py:721)
 
2026-07-17T16:26:08.678Z INFO: Scraping metrics from 0 serving engine(s)
2026-07-17T16:26:10.749Z ERROR: K8s watcher error: (504)
Reason: Timeout: Timeout: Too large resource version: <newer>, current: <current>
 (service_discovery.py:721)
 
...

2026-07-17T16:28:03.203Z ERROR: K8s watcher error: (504)
Reason: Timeout: Timeout: Too large resource version: <newer>, current: <current>
 (service_discovery.py:721)
 ```

### DIFF
--- a/src/tests/test_k8s_service_discovery.py
+++ b/src/tests/test_k8s_service_discovery.py
@@ -9,7 +9,6 @@ from vllm_router.service_discovery import (
     _is_stale_resource_version_error,
 )
 
-
 K8sDiscoveryClass = (
     type[K8sPodIPServiceDiscovery] | type[K8sServiceNameServiceDiscovery]
 )
--- a/src/tests/test_k8s_service_discovery.py
+++ b/src/tests/test_k8s_service_discovery.py
@@ -1,0 +1,184 @@
+from unittest.mock import MagicMock
+
+import pytest
+from kubernetes import client
+
+from vllm_router.service_discovery import (
+    K8sPodIPServiceDiscovery,
+    K8sServiceNameServiceDiscovery,
+    _is_stale_resource_version_error,
+)
+
+
+K8sDiscoveryClass = (
+    type[K8sPodIPServiceDiscovery] | type[K8sServiceNameServiceDiscovery]
+)
+
+
+@pytest.fixture
+def k8s_discovery_setup(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Mock K8s dependencies so service discovery can be unit-tested."""
+    monkeypatch.setattr(
+        "vllm_router.service_discovery.config.load_incluster_config",
+        MagicMock(),
+    )
+    monkeypatch.setattr(
+        "vllm_router.service_discovery.config.load_kube_config",
+        MagicMock(),
+    )
+    monkeypatch.setattr(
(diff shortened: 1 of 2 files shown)


ruizhang0101 (dismissed):
LGTM

loicrouillermonay:
I rebased my branch with master, now PR workflows requires approval from a maintainer 😉

arnaudgeiser:
This PR is important enough since it might affect absolutely everyone using the vLLM Router when the API servers are not in sync with the latest version of ETCD.
The current state of the router makes it highly sensitive to state drift with the current implementation of the Kubernetes watcher.

I cannot speak to the quality of the implementation, but it will definitely improve the router's stability in production workloads where state drift occurs regularly.

loicrouillermonay:
> I rebased my branch with master, now PR workflows requires approval from a maintainer 😉

cc @ruizhang0101

ruizhang0101:
Hi Could you try the PR #1013 and see if this fixed the problem?

loicrouillermonay:
Will do

PR #1004, opened 2026-07-21
fix(router): opt-in --k8s-known-models so scale-to-zero returns 503 not 404 (#1003)

### LINKED ISSUE #1003: Router returns fatal 404 (not retryable 503) for a scaled-to-zero model when the router pod started while the pool was at 0 replicas
`vllm_router` decides 404-vs-503 for a model with no live endpoints using `has_ever_seen_model()` (`src/vllm_router/services/request_service/request.py`):

```python
if not endpoints:
    if not model_ever_existed:
        return JSONResponse(status_code=404, ...)   # fatal
    else:
        return JSONResponse(status_code=503, ...)   # retryable
```

For `K8sPodIPServiceDiscovery`, `has_ever_seen_model()` consults `self.known_models` — a set populated only by pods **this router process** has observed. A router pod that starts (or restarts) while a model's deployment sits at 0 replicas has never observed an engine for it, so it answers a fatal `404 Model not found` for a model that is merely scaled to zero. A sibling router pod started earlier answers a retryable 503 for the same request. With kube-proxy round-robin, ~50% of requests get the fatal answer.

Observed impact (autoscale-to-zero with KEDA): the 404 is self-reinforcing. A fatal 404 fails clients instantly, so they stop sustaining the request-rate metric the autoscaler needs to wake the pool — whereas 503-plus-retry keeps traffic up through the cold start. Combined with a cooldown shorter than the cold start, the model never recovers without operator action.

Numbers from one staging incident: a router pod that started while the pool was at 0 answered 92× 404 / 0× 503 over 20h; its sibling (started when engines existed) answered 466× 503 / 0× 404.

## Suggested fix

Drive the 404/503 split from deployment configuration rather than per-process observation history. Options:

1. Let operators statically declare the deployment's model names (flag or env, e.g. `VLLM_ROUTER_KNOWN_MODELS=modelA,modelB`); `has_ever_seen_model` returns true for declared models. (We run this as a carried patch — happy to PR it.)
2. Have `K8sPodIPServiceDiscovery` watch Deployments/ReplicaSets matching the label selector (which exist at 0 replicas) rather than only Pods.

Happy to contribute either — preference?

Anai-Guo:
## Summary

Fixes #1003 — a router pod that starts (or restarts) while a model's deployment is scaled to **0 replicas** returns a **fatal 404** for that model, instead of a **retryable 503**.

`request_service` decides 404-vs-503 for a model with no live endpoints via `has_ever_seen_model()`:

```python
if not endpoints:
    if not model_ever_existed:
        return JSONResponse(status_code=404, ...)   # fatal
    else:
        return JSONResponse(status_code=503, ...)   # retryable
```

For `K8sPodIPServiceDiscovery`, `has_ever_seen_model()` consults `self.known_models`, a set populated **only by pods this router process has observed**. A router that never saw an engine for a scaled-to-zero model answers a fatal 404, while a sibling pod started earlier answers 503 for the same request. With kube-proxy round-robin ~50% of requests get the fatal answer, and under autoscale-to-zero (KEDA) the 404s are self-reinforcing: clients fail instantly and stop sustaining the request-rate metric the autoscaler needs to wake the pool.

## Change

Implements **option 1** from the issue (the reporter's carried patch): an **opt-in** way to declare a router's known models up front.

- New flag `--k8s-known-models` (comma-separated), defaulting to the `VLLM_ROUTER_KNOWN_MODELS` environment variable.
- `K8sPodIPServiceDiscovery` pre-seeds `known_models` from it at construction, so declared models report as "seen" and return a retryable **503** even before any engine is observed.
- **Default is empty → existing behavior is completely unchanged.** This is purely additive and opt-in.
- Only the **pod-ip** discovery consumes the option; it is popped in `_create_service_discovery` before the service-name variant is constructed, so that path is untouched.

This does not preclude the richer option 2 (watching Deployments/ReplicaSets at 0 replicas) — it's a low-risk escape hatch that operators running scale-to-zero can adopt today.

## Files

| File | What |
|---|---|
| `parsers/parser.py` | add `--k8s-known-models` arg + `VLLM_ROUTER_KNOWN_MODELS` env fallback |
| `app.py` | thread `static_known_models=args.k8s_known_models` into the K8s discovery init |
| `service_discovery.py` | `_parse_static_known_models()` helper; seed `known_models` in pod-ip `__init__`; route the option only to pod-ip |
| `tests/test_static_known_models.py` | unit tests for the parser helper + a seed→`has_ever_seen_model` check (no cluster needed) |

## Test

```bash
pytest src/tests/test_static_known_models.py
```

The construction test bypasses the k8s client/watcher via `__new__`, so it needs no cluster.

### DIFF
--- a/src/tests/test_static_known_models.py
+++ b/src/tests/test_static_known_models.py
@@ -1,0 +1,37 @@
+import pytest
+
+from vllm_router.service_discovery import _parse_static_known_models
+
+
+@pytest.mark.parametrize(
+    "raw, expected",
+    [
+        (None, set()),
+        ("", set()),
+        ("   ", set()),
+        ("modelA", {"modelA"}),
+        ("modelA,modelB", {"modelA", "modelB"}),
+        ("  modelA , modelB ", {"modelA", "modelB"}),
+        ("modelA,,modelB,", {"modelA", "modelB"}),
+        ("modelA,modelA", {"modelA"}),
+    ],
+)
+def test_parse_static_known_models(raw, expected):
+    assert _parse_static_known_models(raw) == expected
+
+
+def test_static_known_models_seed_marks_model_as_seen():
+    """A pod-ip discovery seeded with static models reports them as seen even
+    when no engine has ever been observed (issue #1003, scale-to-zero). The k8s
+    client / watcher thread in __init__ is bypassed via __new__ so no cluster is
+    required."""
+    import threading
+
+    from vllm_router import service_discovery as sd
+
+    instance = sd.K8sPodIPServiceDiscovery.__new__(sd.K8sPodIPServiceDiscovery)
+    instance.known_models = sd._parse_static_known_models("modelA,modelB")
+    instance.known_models_lock = threading.Lock()
+
+    assert instance.has_ever_seen_model("modelA") is True
+    assert instance.has_ever_seen_model("modelC") is False
--- a/src/vllm_router/app.py
+++ b/src/vllm_router/app.py
@@ -234,6 +234,7 @@ def initialize_all(app: FastAPI, args):
             decode_model_labels=args.decode_model_labels,
             watcher_timeout_seconds=args.k8s_watcher_timeout_seconds,
             health_check_timeout_seconds=args.backend_health_check_timeout_seconds,
+            static_known_models=args.k8s_known_models,
         )
 
     elif args.service_discovery == "external-only":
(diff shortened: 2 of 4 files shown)


Anai-Guo:
Addressed the review note in f42c310: `--k8s-known-models` now logs a warning when it is set alongside `service-name` discovery, instead of being silently dropped.

The two failing e2e jobs on the previous run were cluster-side flakes rather than anything from this change — `Timeout waiting for vllm pods to be ready` with `persistentvolumeclaim "vllm-opt125m-storage-claim" is being deleted` / `not found` scheduling warnings, before the router was ever exercised. This push should re-run them.

PR #1005, opened 2026-07-22
Feat(helm): Modify the existing serviceAccount for use by router and lora-controller 

ghkdwlgns612:
## Summary
I wanted to use pre-existing service accounts directly for the router and lora-controller Deployments.
To support this, when serviceAccountName is explicitly set, the chart now skips creating ServiceAccount, Role/ClusterRole, and RoleBinding/ClusterRoleBinding resources.

## Root Cause
The previous Helm templates assumed chart-managed identities and RBAC by default, so they always rendered service-account and RBAC resources when router/lora-controller were enabled. This did not support environments where service accounts and permissions are provisioned externally.

## Changes
- Add routerSpec.serviceAccountName field to values.yaml— when set, use the specified SA for the router deployment
- Add loraController.serviceAccountName field to values.yaml— when set, use the specified SA for the lora-controller deployment
- Skip ServiceAccount creation when a custom serviceAccountName is provided
- Skip Role/RoleBinding creation for router when a custom serviceAccountName is provided
- Skip ClusterRole/ClusterRoleBinding creation for lora-controller when a custom serviceAccountName is provided

### DIFF
--- a/helm/templates/deployment-lora-controller.yaml
+++ b/helm/templates/deployment-lora-controller.yaml
@@ -38,7 +38,7 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-      serviceAccountName: {{ .Release.Name }}-lora-controller
+      serviceAccountName: {{ .Values.loraController.serviceAccountName | default (printf "%s-lora-controller" .Release.Name) }}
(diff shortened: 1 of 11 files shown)


ruizhang0101 (approved):
LGTM

PR #1012, opened 2026-07-23
[Bugfix][Router] Remove ghost endpoint on evicted-worker MODIFIED event

### LINKED ISSUE #1008: bug: router keeps a ghost endpoint after a worker pod is evicted, causing request timeouts
### Describe the bug

This is a real production issue, observed in a live deployment.

In `K8sPodIPServiceDiscovery`, when a worker pod is evicted under node pressure
the kubelet clears its `podIP` and Kubernetes emits a `MODIFIED` event with an
empty IP. The `MODIFIED` branch returns early on an empty IP, so the
already-registered endpoint is never removed. It lingers as a stale ("ghost")
backend: the router keeps routing requests to an address with no worker behind
it, and those requests hang until they time out. The endpoint is only cleared
when the router process restarts.

### To Reproduce

A real node-pressure eviction is hard to force safely on a shared cluster, but
the only code-path difference between an eviction and a normal shutdown is that
the kubelet clears `status.podIP`, producing a `MODIFIED` event with
`podIP=None`. You can reproduce the bug by replicating exactly that status write.

1. Deploy the router with `K8sPodIPServiceDiscovery`. Use a short stats interval
   so the effect is easy to see, e.g. `--engine-stats-interval 2` and
   `--routing-logic roundrobin`.

2. Start one worker pod that the router discovers and registers. Confirm it is
   picked up:

   ```
   Discovered new serving engine <pod> at <POD_IP>
   Scraping metrics from 1 serving engine(s)
   ```

3. Simulate the eviction by patching the pod's status subresource to clear its
   IP (this is exactly what the kubelet does on eviction):

   ```bash
   kubectl proxy --port=8801 &
   curl -X PATCH \
     "http://127.0.0.1:8801/api/v1/namespaces/<ns>/pods/<pod>/status" \
     -H "Content-Type: application/merge-patch+json" \
     -d '{"status":{"phase":"Failed","reason":"Evicted",
           "podIP":null,"podIPs":null,
           "containerStatuses":[{"name":"<container>","ready":false,
             "state":{"terminated":{"exitCode":137,
               "reason":"ContainerStatusUnknown"}}}]}}'
   ```

   (On kubelet/kubectl versions without `--subresource=status`, the proxy +
   PATCH approach above is the reliable way to write the status subresource.)

4. Observe the bug: the router keeps `<POD_IP>` in its routable set and keeps
   polling it, even though no pod holds that IP anymore:

   ```
   ERROR Failed to scrape metrics from http://<POD_IP>:8000:
         ConnectTimeoutError(... connect timeout=2)
   Server: http://<POD_IP>:8000        <- still being routed to
   ```

   The realtime routable count stays at `1` instead of dropping to `0`. The
   ghost endpoint persists until the router process is restarted.

> **Note on observability:** Use the live
> `Scraping metrics from N serving engine(s)` count, the
> `Failed to scrape ... ConnectTimeout` lines, and the
> `Serving engine X is deleted` log to judge whether the endpoint was actually
> removed.

### Expected behavior

After the `podIP` is cleared (evicted pod), the router should remove that
endpoint within about one watch cycle — the same way it handles a `DELETED`
event — logging `Serving engine X is deleted`, dropping the realtime routable
count to `0`, and producing no further `Failed to scrape` polling of the dead
address. No router restart should be required.

### Additional context

Encountered in a production environment, where evicted workers left ghost
endpoints that kept receiving traffic until the router was restarted.

The empty IP on a `MODIFIED` event is ambiguous: a *Pending* pod has no IP yet
(should be skipped), while an *Evicted* pod has had its `podIP` cleared (its
endpoint must be removed). The current code treats both the same and skips
removal.

jiawen7777:
When a worker pod is evicted under node pressure, the kubelet clears its podIP and Kubernetes emits a MODIFIED event with engine_ip=None. On current main the MODIFIED branch returns early on a None IP, so the already-registered endpoint lingers as a stale ("ghost") backend. The router keeps routing requests to that address even though no worker is behind it, and those requests hang until they time out. The endpoint is only cleared when the router process restarts.

An empty IP on MODIFIED is ambiguous: a Pending pod has no IP yet (skip it), but an Evicted pod has had its podIP cleared and its endpoint must be removed. Remove the registered engine on this path, mirroring the DELETED handling, and log a WARNING so the removal is observable. This fix has been verified to resolve the problem in real deployment.

Scope: K8sPodIPServiceDiscovery only; ADDED and graceful-drain paths unchanged; no routing changes. Adds 5 regression tests.

FILL IN THE PR DESCRIPTION HERE

FIX #1008

---

<details>

<ul>
</ul>

<h3>Code Quality</h3>

<ul>
</ul>

, Shaoting-Feng or ApostaC.


ruizhang0101 (approved):
LGTM

jiawen7777:
Hi @ruizhang0101 thanks for reviewing!
 
I noticed the only failing job is `Two-Pods-Minimal-Example`, which seems to have timed out after 3 minutes on the "Validate the installation and send query to the stack" step. My understanding is that this might be a startup timeout rather than something caused by this change, 

Would it be possible to re-run that job to see if it passes on a second try?  Thanks for your time~

PR #1017, opened 2026-07-28
[Router][Bugfix] Return 400 for malformed request payloads

tiebreaker4869:
## Summary

This PR returns HTTP 400 responses for invalid client request payloads instead of allowing parsing errors to propagate as HTTP 500 responses.

Specifically, it:
- returns HTTP 400 for malformed or undecodable JSON request bodies;
- returns HTTP 400 when the top-level JSON value is not an object;
- returns HTTP 400 when the transcription form data is invalid;
- adds regression tests for these request-validation cases.

## Testing
- pre-commit check and unittests green
- Local HTTP verification:
    - malformed JSON returns 400
    - non-object JSON returns 400
    -  invalid transcription form data returns 400

---

<details>

<ul>
</ul>

<h3>Code Quality</h3>

<ul>
</ul>

, Shaoting-Feng or ApostaC.


src/vllm_router/services/request_service/request.py
```diff
@@ -1139,6 +1151,11 @@ async def route_general_transcriptions(
             status_code=400,
             content={"error": f"Invalid request: missing '{e.args[0]}' in form data."},
         )
+    except (TypeError, ValueError):
+        return JSONResponse(
```
ruizhang0101, line 1160: Could you return request id as well?

src/vllm_router/services/request_service/request.py
```diff
@@ -1139,6 +1151,11 @@ async def route_general_transcriptions(
...
+            status_code=400,
+            content={"error": "Invalid request: 'temperature' must be a number."},
```
ruizhang0101: Not only temperature will have these error here. The request.form might also raise that. Consider using a broader scope message.

ruizhang0101 (changes requested):
otherwise LGTM

ruizhang0101 (dismissed):
LGTM

lfsun02:
looks like there are some timeout issues in checks. Could you take a look when you have a chance? Thanks! @ruizhang0101

ErenAta16 (commented):
Direction is right, and the `isinstance(request_json, dict)` guard is the part I would have most expected to be forgotten. It catches every non-object JSON document, which the downstream code would otherwise index into:

```
b'[]'    -> list      b'5'     -> int
b'"x"'   -> str       b'null'  -> NoneType
b'{}'    -> dict      (the only one that proceeds)
```

Threading `request_id` into the transcription path so the 400 carries `X-Request-Id` is a good catch too, since a client correlating by that header would otherwise lose exactly the failed requests.

**The except clause is one exception short.** `json.loads` does not only raise `JSONDecodeError` and `UnicodeDecodeError`. A deeply nested document exhausts the interpreter stack, and `RecursionError` is not a subclass of either:

```
depth  1000 : parsed ok        (2 KB payload)
depth 10000 : RecursionError   (20 KB payload)
depth 100000: RecursionError   (200 KB payload)

issubclass(RecursionError, (json.JSONDecodeError, UnicodeDecodeError))  ->  False
```

So `b"[" * 10000 + b"]" * 10000`, a 20 KB body that any client can send, still lands as a 500 after this PR rather than the 400 the PR is introducing. That is not a new problem, but it sits squarely inside what this change is trying to fix, and it is cheap to trigger, so I would fold it in rather than leave a second PR's worth of the same bug behind.

Widening to `except (json.JSONDecodeError, UnicodeDecodeError, RecursionError)` covers it. `except ValueError` would too, since `JSONDecodeError` subclasses it, but that also swallows things you probably want to see, so the explicit tuple reads better.

Worth noting the message can stay the same. From the client's point of view a body the parser cannot handle is a malformed body either way, and "must be valid JSON" is accurate for both.

Two smaller things on the tests, since a test file is the substance of the diff here:

The malformed-payload cases would be stronger with a nesting case alongside the bad-syntax ones, and it would fail today, which is the property that keeps it honest.

`_json_request` builds a `SimpleNamespace` rather than a real `Request`. That is fine for these paths, but it means the test cannot catch a regression where the handler starts touching an attribute the namespace does not define. Worth a comment saying so, otherwise the next person extends the handler and gets an `AttributeError` from the test rather than a meaningful failure.

ruizhang0101 (approved):
LGTM

lfsun02:
> LGTM

looks like the gatekeeper does not do deduplication on workflow runs of the same check, and treats previous cancelled runs as failed even when new runs pass.

https://github.com/vllm-project/production-stack/actions/runs/30885907838/job/92122841550?pr=1017

PR #1020, opened 2026-07-28
fix: await router.route_request and pass request_json in proxy_multipart_request [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

### LINKED ISSUE #1019: bug: SessionRouter.route_request() missing `request_json` argument in proxy_multipart_request
### Describe the bug

When routing audio transcription requests (e.g. `/v1/audio/transcriptions`) to a Whisper model using `--routing-logic session`, the router raises:

```
TypeError: SessionRouter.router_request() missing 1 required positional argument: 'request_json'
```

This is caused by `proxy_multipart_request` in `src/vllm_router/services/request_service/request.py` calling `router.route_request(...)` **without** the required `request_json` argument, and also calling it **without `await`** even though `SessionRouter.route_request` is an `async` method.

Affected code (around line 1245):

```python
# BUGGY — missing request_json, and not awaited
chosen_url = router.route_request(
    endpoints,
    engine_stats,
    request_stats,
    request,
)
```

By contrast, the regular JSON request path in `route_general_request` correctly calls:

```python
server_url = await request.app.state.router.route_request(
    endpoints, engine_stats, request_stats, request, request_json
)
```

The fix is to pass `{}` as `request_json` (multipart/form-data requests have no JSON body) and add `await`:

```python
# FIXED
chosen_url = await router.route_request(
    endpoints,
    engine_stats,
    request_stats,
    request,
    {},  # request_json — no JSON body for multipart/form-data
)
```

Note: if the router is NOT a `SessionRouter` / `KvawareRouter` / `PrefixAwareRouter`, the call should remain synchronous and without `request_json` (matching the else-branch pattern in `route_general_request`). The call site should mirror the same `isinstance` guard used elsewhere.

### To Reproduce

1. Start a vLLM backend serving a Whisper model with `--task transcription`:
   ```bash
   vllm serve --task transcription openai/whisper-large-v3 --host 0.0.0.0 --port 8002
   ```
2. Start the router with **session** routing logic:
   ```bash
   python3 -m vllm_router.app \
     --port 8000 \
     --service-discovery static \
     --static-backends http://localhost:8002 \
     --static-models openai/whisper-large-v3 \
     --static-model-types transcription \
     --routing-logic session \
     --session-key x-user-id
   ```
3. Send a transcription request:
   ```bash
   curl http://localhost:8000/v1/audio/transcriptions \
     -F 'file=@audio.wav;type=audio/wav' \
     -F 'model=openai/whisper-large-v3'
   ```
4. Observe the error:
   ```
   TypeError: SessionRouter.router_request() missing 1 required positional argument: 'request_json'
   ```

### Expected behavior

The transcription request should be successfully routed to the Whisper backend using `SessionRouter` (or any other async-capable router) without raising a `TypeError`.

### Additional context

- The bug exists in `proxy_multipart_request` in `src/vllm_router/services/request_service/request.py` (around line 1245).
- The same code path is used by `route_general_transcriptions` (`/v1/audio/transcriptions`) and `route_image_edit_request` (`/v1/images/edits`).
- `roundrobin` routing is **not** affected because `RoundRobinRouter.route_request` is synchronous and does not require `request_json`.
- Only routers that override the async signature (`SessionRouter`, `KvawareRouter`, `PrefixAwareRouter`) are affected.
- Fix PR: add an `isinstance` guard (matching the pattern in `route_general_request`) and pass `{}` as `request_json` for the async router branch.

waterWang:
Fixes #1019

## What does this PR do?

Fixes a bug in `proxy_multipart_request` where `SessionRouter.route_request()` was called without the required `request_json` argument, and without `await`.

## Root Cause

The `proxy_multipart_request` function (used for audio transcription and image edit requests) called `router.route_request()` synchronously without `request_json`:

```python
chosen_url = router.route_request(
    endpoints,
    engine_stats,
    request_stats,
    request,
)
```

However, `SessionRouter`, `KvawareRouter`, and `PrefixAwareRouter` all have an async `route_request` method that requires a `request_json` argument. The regular `route_general_request` function already handles this correctly with an `isinstance` guard.

## Fix

Mirrors the pattern used in `route_general_request`:
- If the router is a `SessionRouter`, `KvawareRouter`, or `PrefixAwareRouter`, call `await router.route_request(...)` with `{}` as `request_json` (multipart/form-data requests have no JSON body).
- Otherwise, call `router.route_request(...)` synchronously without `request_json`.

## Testing

- The fix follows the exact same pattern as `route_general_request` (lines 561-570), which is already well-tested.
- Only affects multipart routes (`/v1/audio/transcriptions`, `/v1/images/edits`) when using session/kv-aware/prefix-aware routing.
- `roundrobin` routing is unaffected.

### DIFF
--- a/src/vllm_router/services/request_service/request.py
+++ b/src/vllm_router/services/request_service/request.py
@@ -1242,12 +1242,21 @@ async def proxy_multipart_request(
     request_stats = request_stats_monitor.get_request_stats(time.time())
 
     # pick one using the router's configured logic (roundrobin, least-loaded, etc.)
-    chosen_url = router.route_request(
-        endpoints,
-        engine_stats,
-        request_stats,
-        request,
-    )
+    if isinstance(router, (KvawareRouter, PrefixAwareRouter, SessionRouter)):
+        chosen_url = await router.route_request(
+            endpoints,
+            engine_stats,
+            request_stats,
+            request,
+            {},
+        )
+    else:
+        chosen_url = router.route_request(
+            endpoints,
+            engine_stats,
+            request_stats,
+            request,
+        )
     logger.info(
         "Proxying multi-part form request for model %s to %s", model, chosen_url
     )


ruizhang0101:
Duplicate with #1021

ErenAta16 (commented):
Confirmed the crash rather than inferring it from the signature. Reproducing the call exactly as `proxy_multipart_request` makes it on `main`:

```
SessionRouter.route_request signature:
    (self, endpoints, engine_stats, request_stats, request, request_json) -> str
is coroutine function: True

sr.route_request(endpoints, engine_stats, request_stats, request)
    TypeError: SessionRouter.route_request() missing 1 required positional argument: 'request_json'
```

So it fails at the call, before the missing `await` even becomes relevant. Every multipart request, audio transcription and image edit, is a hard 500 whenever the router is one of the three async ones. `RoundRobinRouter.route_request` is not a coroutine and takes four arguments, which is why the plain path never noticed.

The `isinstance` dispatch matches what `services/request_service/request.py` already does around line 561, so this keeps the two call sites in step rather than inventing a second convention. That is the right call, though it does mean the tuple `(KvawareRouter, PrefixAwareRouter, SessionRouter)` now appears in two places and will need updating in both when a fourth async router lands. A small helper, or an `is_async` flag on `RoutingInterface`, would remove that trap. Not a blocker, and probably better as a follow-up than bundled here.

**One behavioural consequence worth writing down.** Passing `{}` is correct in the sense that `extract_session_id` calls `request_json.get(...)` and would raise on `None`, so it has to be a mapping. But it means session extraction on these endpoints is header-only:

```python
val = request.headers.get(session_key)
return val if val is not None else request_json.get(session_key, None)
```

For a JSON endpoint the body is the fallback. For multipart there is no JSON body to read, so `{}` is honest, but a client that puts its session id in a multipart *form field* rather than a header will now silently get QPS routing instead of session affinity. Silently, because a missing session id is a valid state that falls through to `_qps_routing`, not an error.

If transcription is expected to participate in session affinity at all, parsing the session key out of the form data and passing `{session_key: value}` would close that. If it is not expected to, a line in the docstring saying multipart requests are header-only would save someone the debugging session. Either way I would not hold the PR for it, since header-only affinity is strictly better than a 500.

Fix looks right to me.

Issue #1022, opened 2026-07-28
refactor: standardize route_request signature across all routers


shernshiou:
## Background

While fixing #1019 (missing `request_json` and missing `await` in `proxy_multipart_request`), it became clear that the `route_request` method is inconsistent across router implementations:

| Router | Async? | Accepts `request_json`? |
|---|---|---|
| `RoundRobinRouter` | ❌ | ❌ |
| `SessionRouter` | ✅ | ✅ |
| `KvawareRouter` | ✅ | ✅ |
| `PrefixAwareRouter` | ✅ | ✅ |
| `DisaggregatedPrefillRouter` | ❌ | ✅ |
| `DisaggregatedPrefillOrchestratedRouter` | ✅ | ✅ |

This forces every call site (`route_general_request`, `proxy_multipart_request`) to maintain `isinstance` guards to dispatch the right variant — a fragile pattern that will break silently whenever a new router is added.

## Proposed Fix

Standardize the base class `RoutingInterface.route_request` to:

```python
async def route_request(
    self,
    endpoints: List[EndpointInfo],
    engine_stats: Dict[str, EngineStats],
    request_stats: Dict[str, RequestStats],
    request: Request,
    request_json: dict = {},
) -> str:
```

- Make all routers `async def route_request` (sync routers just don't `await` internally)
- Add `request_json: dict = {}` default to routers that don't use it (`RoundRobinRouter`)
- Update all call sites to always `await` and always pass `request_json`
- Remove all `isinstance` routing guards from `request.py`

## Scope

- `src/vllm_router/routers/routing_logic.py` — 6 router classes
- `src/vllm_router/services/request_service/request.py` — ~4 call sites
- Relevant tests

Relates to #1019

PR #1023, opened 2026-07-28
[Router] Standardize route_request signature across all routers

Mohith26:
Implements #1022, proposed by @shernshiou.

`route_request` was inconsistent across the 6 router implementations (some sync, some async, some without `request_json`), which forced isinstance dispatch guards at every call site and would break silently whenever a new router is added.

Changes:
- `RoutingInterface.route_request` is now `async def` and takes `request_json`; all 6 routers share the signature. Sync routers just do not await internally; no routing behavior changed.
- `route_general_request` (initial + failover call sites) and `proxy_multipart_request` always await and always pass `request_json`; the `(KvawareRouter, PrefixAwareRouter, SessionRouter)` isinstance guards are gone. The `DisaggregatedPrefill*` isinstance checks stay because they dispatch to different request handlers, not `route_request` variants.
- One deviation from the issue text: `request_json: Optional[Dict] = None` (normalized inside) instead of the mutable `dict = {}` default, matching the repo's existing typing style and avoiding the shared-default hazard. Happy to change if you prefer the literal proposal.

Tests: new `test_route_request_signature.py` (16 tests) checks every router's signature plus behavioral probes proving the call sites really await and forward the body (via a router subclass that is not in any old isinstance whitelist). Existing roundrobin/auth-header tests updated for async. Full suite: 186 passed, 0 failed (baseline on main was 170 passed, 0 failed). black/isort/ruff clean.

Relates to #1019: this standardizes the surface that made that bug possible. PRs #1020/#1021 fix the immediate call-site bug; this PR is complementary and rebases cleanly in either order.

### DIFF
--- a/src/tests/test_request_auth_headers.py
+++ b/src/tests/test_request_auth_headers.py
@@ -32,7 +32,7 @@ def _build_request(headers=None):
     state.callbacks = None
     state.request_stats_monitor.get_request_stats.return_value = {}
     state.engine_stats_scraper.get_engine_stats.return_value = {}
-    state.router.route_request.return_value = "http://whisper-engine"
+    state.router.route_request = AsyncMock(return_value="http://whisper-engine")
 
     request = MagicMock()
     request.headers = headers or {}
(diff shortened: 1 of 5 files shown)


ruizhang0101:
Hi, could you resolve the conflict when you have time?

PR #1024, opened 2026-07-28
[CI/Build] Move test scripts and assets from .github/ to tests/ (part of #531)

ighutake-debug:
Implements item 1 ("Move the scripts and assets from `.github/` to `tests/") of the CI refactor plan agreed in #531 ([comment](https://github.com/vllm-project/production-stack/issues/531#issuecomment-5098069379)).

Part of #531 — this PR covers only item 1 of 4; follow-up PRs will address the fake-server, job-consolidation, and Helm de-duplication items, so merging this should NOT close the issue.

## What changed

Moved 15 loose CI files out of `.github/`:
- `tests/scripts/`: `curl-02-two-pods.sh`, `curl-04-multiple-models.sh`, `curl-05-secure-vllm.sh`, `curl-06-monitoring.sh`, `port-forward.sh`
- `tests/assets/`: `values-01/04/05/06/07/08/09/10/11-*.yaml`, `template-chatml.jinja`

Updated every referencing call site:
- `.github/workflows/functionality-helm-chart.yml` — `helm install -f` values paths + `port-forward.sh` invocations
- `.github/workflows/router-e2e-test.yml` — `--chat-template` path for the static-discovery backends
- `tests/e2e/run-k8s-routing-test.sh` — `HELM_VALUES_FILE` paths (10 references)
- `tests/scripts/port-forward.sh` — path to the curl scripts it dispatches

Also added `tests/assets/**` and `tests/scripts/**` to the `paths` filters of both workflows so changes to the moved files keep triggering CI.

No functional changes — pure move + reference updates (all 14 content files are 100% renames; only `port-forward.sh` has a 1-line path edit).

## Validation done locally

- `grep` sweep: zero remaining references to the old `.github/` paths
- `bash -n` passes on all moved/edited shell scripts
- YAML parse passes on both workflows and all moved values files
- `pre-commit run` on all changed files: actionlint, check-yaml, end-of-file, trailing-whitespace, codespell all pass

The self-hosted runner jobs (`functionality-helm-chart.yml`, `router-e2e-test.yml`) will exercise the moved scripts end-to-end on this PR's CI run.

**BEFORE SUBMITTING, PLEASE READ THE CHECKLIST BELOW AND FILL IN THE DESCRIPTION ABOVE**

---

- [x] Make sure the code changes pass the [pre-commit](https://github.com/vllm-project/production-stack/blob/main/CONTRIBUTING.md) checks.
- [x] Sign-off your commit by using <code>-s</code> when doing <code>git commit</code>
[...]

PR #1025, opened 2026-07-29
[Router][Bugfix] Fix KvawareRouter event-loop blocking during tokenization (#1016)

### LINKED ISSUE #1016: KvawareRouter blocks the event loop per request (sync tokenizer fetch + /tokenize POST + controller lookup) — health probes starve, router CrashLoops under load
## Describe the bug

`KvawareRouter.route_request` performs blocking I/O on the uvicorn event loop for every routed request, which starves the router's own `/health` endpoint. Under Kubernetes liveness/readiness probes this makes the router CrashLoop as soon as it receives sustained traffic — we observed 25 liveness-kill restarts in ~1h on a single-replica kvaware router receiving one ~30K-token request every 15s, with zero requests completing.

Three blocking calls in `src/vllm_router/routers/routing_logic.py` (`KvawareRouter.route_request`):

1. `self.tokenizer = AutoTokenizer.from_pretrained(endpoints[0].model_names[0])` — a synchronous network fetch on the event loop. When the served model name is an alias (a common `--served-model-name` pattern, e.g. `default`), this can never succeed; in a locked-down cluster (egress policy blocking huggingface.co) it blocks until the connection attempt times out — every request, since `self.tokenizer` is only assigned on success.
2. The fallback `requests.post(endpoints[0].url + "/tokenize", ..., timeout=10)` — a synchronous HTTP POST of the **full prompt** (seconds of wall-clock for long prompts), on the event loop.
3. The controller lookup via `query_manager` → `handle_orchestration_message`, which blocks the calling thread while the ZMQ round-trip completes on the controller's own loop.

While any of these run, the event loop cannot answer `/health`, so the kubelet's probe times out ("context deadline exceeded"), readiness drops (the Service refuses new connections), and after the liveness threshold the container is killed mid-flight. The failure clears instantly when traffic stops, confirming the per-request wedge.

## To Reproduce

1. Deploy vllm-router with `--routing-logic kvaware` + the three `--lmcache-controller-*` ports, fronting one vLLM pod running an LMCache worker (`LMCACHE_ENABLE_CONTROLLER=True`, registered via the router-hosted controller). Standard K8s liveness/readiness probes on `/health` (2s timeout).
2. Ensure the router pod cannot reach huggingface.co (default-deny egress, common in production), and serve the model under an alias via `--served-model-name`.
3. Send one long-prompt (~30K tokens) chat completion every ~15s.
4. Observe: `/health` probe timeouts begin with the first routed request; the router flaps NotReady, then CrashLoops (`Killing ... failed liveness probe`). `vllm-router` logs show normal engine scraping between wedges (those run on background threads, masking the stalled loop).

## Expected behavior

Routing-decision work should not block the event loop: tokenizer resolution off-loop (`asyncio.to_thread` / executor) and cached across requests including the failure case, the `/tokenize` fallback via the async HTTP client, and the controller lookup awaited rather than blocked on. `/health` should stay responsive regardless of routing-path latency.

## Environment

- production-stack `vllm-router` @ `7c273ab40b671a13fb5a7389f085f336a2c907f4`
- LMCache 0.5.2 (router-side controller + worker-side `LMCacheMPConnector`)
- vLLM 0.25.1 workers, K8s (EKS), single-replica router, probes on `/health`

---

_This issue was researched and written by Claude (an AI assistant) operating on behalf of @HipHoff, who reviewed it before filing. Happy to provide full pod logs/events or test a patch._

ighutake-debug:
Fixes #1016.

`KvawareRouter.route_request` performed blocking I/O on the uvicorn event loop for every routed request, starving `/health` probes and CrashLooping the router under sustained traffic (25 liveness-kills/hour in the reporter's deployment). Thanks @HipHoff for the downstream patch shape and @waynehacking8 for the lightweight repro — the fix follows both inputs.

## What changed (src/vllm_router/routers/routing_logic.py)

1. **Tokenizer load off the loop** — `AutoTokenizer.from_pretrained` (network-bound HF hub fetch, can hang for seconds with blocked egress) now runs via `asyncio.to_thread`.
2. **Fallback `/tokenize` POST off the loop** — the `requests.post` carrying the full prompt (10s timeout) now runs via `asyncio.to_thread`.
3. **Failed loads are cached** — new `_tokenizer_load_failed` flag: a doomed load (alias `--served-model-name`, egress-blocked huggingface.co) is attempted once, then the router goes straight to the remote `/tokenize` fallback. Previously only *successful* loads were cached, so doomed loads were retried — and re-blocked the loop — on every request.

Behavior is otherwise identical: local-first tokenization, remote fallback on any failure, errors from the fallback still bubble.

## Deliberately NOT changed: `query_manager`

The issue lists the controller lookup as a third blocking call. On current lmcache, `handle_orchestration_message` → [`kv_controller.lookup`](https://github.com/LMCache/LMCache/blob/main/lmcache/v1/cache_controller/controllers/kv_controller.py#L388) is an **in-memory registry read** (no ZMQ round-trip), so awaiting it does not block the loop. Moving it to an ad-hoc thread would risk racing the controller's own registry state, so it's left alone. Happy to revisit if maintainers see blocking there in their deployments.

## Tests

New `src/tests/test_kvaware_router_event_loop.py` (4 tests, all failing pre-fix except the characterization guard):

- `test_tokenizer_load_does_not_block_event_loop` — probe starved at 1 tick pre-fix
- `test_remote_tokenize_fallback_does_not_block_event_loop` — probe starved at 1 tick pre-fix
- `test_failed_tokenizer_load_is_not_retried_every_request` — `from_pretrained` called 3× pre-fix, 1× post-fix
- `test_local_tokenizer_loaded_once_and_reused` — unchanged behavior guard

Full `src/tests` suite: **116 passed**. `pre-commit` on changed files: black, isort, ruff, codespell all pass.

---

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/tests/test_kvaware_router_event_loop.py
+++ b/src/tests/test_kvaware_router_event_loop.py
@@ -1,0 +1,182 @@
+# Copyright 2024-2025 The vLLM Production Stack Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Regression tests for https://github.com/vllm-project/production-stack/issues/1016
+
+KvawareRouter.route_request performed blocking I/O (tokenizer download,
+fallback /tokenize HTTP POST) directly on the uvicorn event loop, starving
+/health probes and CrashLooping the router under sustained traffic.
+"""
+
+import asyncio
+import time
+from types import SimpleNamespace
+from typing import Any, Dict
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+from uhashring import HashRing
+
+from vllm_router.routers import routing_logic
+from vllm_router.routers.routing_logic import KvawareRouter
+
+
+@pytest.fixture(autouse=True)
+def stub_lmcache_names(monkeypatch):
+    """lmcache is an optional runtime dep; provide the message names the
+    router constructs (they only reach the mocked query_manager)."""
+    monkeypatch.setattr(routing_logic, "LookupMsg", MagicMock(), raising=False)
+    monkeypatch.setattr(routing_logic, "QueryInstMsg", MagicMock(), raising=False)
+
+
+class EndpointInfo:
+    def __init__(self, url: str, model_names=None):
+        self.url = url
+        self.model_names = model_names or ["test-model"]
+
+
+class Request:
+    def __init__(self, headers: Dict[str, str], body: Dict[str, Any] = None):
+        self.headers = headers
+        self.body = body
+
+
+class FakeTokenizer:
+    def encode(self, text):
+        return [1, 2, 3]
+
+
+def make_router() -> KvawareRouter:
+    """KvawareRouter without __init__ (avoids LMCache controller setup)."""
+    router = KvawareRouter.__new__(KvawareRouter)
+    router.tokenizer = None
+    router._tokenizer_load_failed = False
+    router.session_key = None
+    router.threshold = 2000
+    router.hash_ring = HashRing()
+    router.instance_id_to_ip = {}
+    # No KV layout hits: fall through to the QPS/session fallback path,
+    # which returns endpoints[0].url when request_stats is empty.
+    router.query_manager = AsyncMock(return_value=SimpleNamespace(layout_info={}))
+    return router
+
+
+async def run_with_health_probe(coro, probe_interval=0.05):
+    """Run coro alongside a fake /health probe; return probe ticks at coro end."""
+    ticks = 0
+
+    async def health_probe():
+        nonlocal ticks
+        for _ in range(20):
+            ticks += 1
+            await asyncio.sleep(probe_interval)
+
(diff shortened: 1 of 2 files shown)


waynehacking8:
I tested e104ed573cd3177a8c6c5adbee3398f3c14b1d86 on NVIDIA RTX PRO 6000 Blackwell, SM120, 96GB with the exact #1016 restricted egress workload. Health recovered completely: 177/177 samples returned 200, p95 was 0.007987s, maximum was 0.014888s, and the trial had no probe failures or router restarts; backend `/tokenize`, controller lookup, KV-aware routing, generation, and LMCache hits all ran. Only 3/12 clients completed within 14s; six requests arriving before the first roughly 80s `from_pretrained` failure finished each started another load, confirming the concurrent load race, while the next three reached fallback quickly but timed out during the backend first-request LMCache post-initialization.

ighutake-debug:
Friendly ping for review when you have a moment.

**CI note:** The `Two-Pods-Minimal-Example` / `All Checks Passed` failures on the earlier run were unrelated to the router change — the branch predated #1024 (moved values to `tests/assets/`) and #1027 (checkout on self-hosted Helm jobs), so the Helm job was still referencing `.github/values-01-2pods-minimal-example.yaml`. I've rebased onto current `main` to pick those up; CI should rerun cleanly.

**Scope:** This PR only touches `routing_logic.py` and adds `test_kvaware_router_event_loop.py` (4 regression tests). Python Tests, pre-commit, and the router e2e jobs were already green.

**Validation:** Thanks again to @waynehacking8 for the Blackwell restricted-egress repro on commit `e104ed5` — health probes recovered 177/177 after the fix.

Happy to adjust anything if maintainers prefer a different approach for the concurrent `from_pretrained` race he noted.

ighutake-debug:
Closing — no longer pursuing this contribution. Thanks for the review consideration.

PR #1026, opened 2026-07-31
[Doc] Fix references to non-existent tutorial values files

latent-9:
Three tutorials (and their `docs/source/use_cases/` mirrors) reference values files that do not exist under `tutorials/assets/`, so the documented `helm install -f ...` commands fail verbatim with `Error: open ...: no such file or directory`.

| Documented (wrong) | Actual file in `tutorials/assets/` |
|---|---|
| `values-08-tool-enabled.yaml` | `values-13-tool-enabled.yaml` |
| `values-11-secure-vllm-serve.yaml` | `values-11-secure-vllm.yaml` |
| `values-06-remote-shared-storage.yaml` | `values-06-shared-storage.yaml` |

Only the right-hand names exist in the repo. Notably, the secure-serve tutorial (`11`) and shared-storage tutorial (`06`) already use the correct name in Step 1, and only a later command / markdown link uses the wrong one — so this is a clear internal inconsistency.

Files touched: `tutorials/13-tool-enabled-installation.md`, `tutorials/11-secure-vllm-serve.md`, `tutorials/06-remote-shared-kv-cache.md`, `docs/source/use_cases/tool-enabled-installation.rst`, `docs/source/use_cases/sharing-kv-cache.rst`. Docs-only.

### DIFF
--- a/docs/source/use_cases/sharing-kv-cache.rst
+++ b/docs/source/use_cases/sharing-kv-cache.rst
@@ -28,7 +28,7 @@ Prerequisites
 Step 1: Configuring Remote KV Cache Storage
 --------------------------------------------
 
-Locate the file `tutorials/assets/values-06-remote-shared-storage.yaml <https://github.com/vllm-project/production-stack/blob/main/tutorials/assets/values-06-remote-shared-storage.yaml>`_ with the following content:
+Locate the file `tutorials/assets/values-06-shared-storage.yaml <https://github.com/vllm-project/production-stack/blob/main/tutorials/assets/values-06-shared-storage.yaml>`_ with the following content:
 
 .. code-block:: yaml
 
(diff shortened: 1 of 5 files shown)


ErenAta16 (commented):
All three corrections check out against the tree, and the old names are not just moved, they do not exist anywhere in the repository:

```
values-06-remote-shared-storage.yaml   absent   ->  values-06-shared-storage.yaml   present
values-08-tool-enabled.yaml            absent   ->  values-13-tool-enabled.yaml     present
values-11-secure-vllm-serve.yaml       absent   ->  values-11-secure-vllm.yaml      present
```

So anyone following those tutorials literally was getting a "file not found" from `helm install`, which is a bad first experience in exactly the docs that are meant to onboard people.

Since the obvious question on a fix like this is whether it caught all of them, I swept the whole repo rather than spot-checking: 29 distinct `values-*.yaml` filenames are referenced across the `.md` and `.rst` files, and after this PR every one of them resolves to a file that exists. Two that look missing if you only search `tutorials/assets` are fine, `values-01-offload-kv-cache-local-disk.yaml` lives under `tutorials/assets/gke-example/` and `values-example.yaml` under `helm/`, both referenced with paths that match where they actually are.

Nothing left over. Looks complete to me.

ruizhang0101 (approved):
LGTM

PR #1069, opened 2026-08-29
[Bugfix][Router] Run KvawareRouter /tokenize fallback in executor to avoid blocking event loop

Asthenia0412:
## Description

`KvawareRouter.route_request()` uses a synchronous `requests.post` call for the remote `/tokenize` fallback directly on the event loop. Under load, this blocking call stalls the event loop and can delay health checks and other concurrent requests.

The same pattern was already fixed in `tokenize_prompt()` by running the blocking HTTP call via `asyncio.get_running_loop().run_in_executor()`. This PR applies the same fix to the `route_request()` fallback for consistency.

## Change

`src/vllm_router/routers/routing_logic.py`

```python
# before
body = requests.post(remote_url, headers=headers, json=data, timeout=10).json()
token_ids = body["tokens"]

# after
loop = asyncio.get_running_loop()
response = await loop.run_in_executor(
    None,
    lambda: requests.post(remote_url, headers=headers, json=data, timeout=10),
)
token_ids = response.json()["tokens"]
```

## Testing

- Syntax verified with `ast.parse`
- No behavior change; only moves the blocking call off the event loop

### DIFF
--- a/src/vllm_router/routers/routing_logic.py
+++ b/src/vllm_router/routers/routing_logic.py
@@ -402,10 +402,16 @@ class KvawareRouter(RoutingInterface):
                 "model": endpoints[0].model_names[0],
                 "prompt": request_json.get("prompt", ""),
             }
-            body = requests.post(
-                remote_url, headers=headers, json=data, timeout=10
-            ).json()
-            token_ids = body["tokens"]
+            # Run the blocking HTTP call in an executor so it does not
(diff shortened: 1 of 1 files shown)


Asthenia0412:
Thanks for the review! I updated the fallback to use the shared `aiohttp` client session from `request.app.state.aiohttp_client_wrapper()` instead of `requests.post` in an executor. This reuses TCP connections and keeps the call off the event loop. Could you take another look?

PR #1070, opened 2026-09-01
[Bugfix][Router] Default transcription language to None (auto-detect)

WaelRabah:
## Summary
`route_general_transcriptions` was defaulting the transcription `language` to `"en"` when the client did not supply one. This forced the backend to transcribe in English even when the model could auto-detect the language.

vLLM's own default for `language` is `None`, which leaves language detection to the model. This change aligns the router with that behavior.

## Changes
- `src/vllm_router/services/request_service/request.py`
  - Default `language` to `None` (`form.get("language")` instead of `form.get("language", "en")`).
  - Only include the `language` field in the outbound multipart form data when it is set. When absent, `aiohttp.FormData` would otherwise encode `None` as the literal string `"None"` and send it to the backend as an invalid language.
- `src/tests/test_multipart_proxy.py`
  - Update the audio transcription tests to assert the field is absent/`None` by default.

## Testing
`src/tests/test_multipart_proxy.py` — 4 passed.

### DIFF
--- a/src/tests/test_multipart_proxy.py
+++ b/src/tests/test_multipart_proxy.py
@@ -165,7 +165,8 @@ async def test_audio_transcription_preserves_plain_text_response():
         received.update(
             model=form["model"],
             response_format=form["response_format"],
-            language=form["language"],
+            # absent by default so the backend auto-detects; None documents that
+            language=form.get("language"),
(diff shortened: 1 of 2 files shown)


ruizhang0101 (approved):
LGTM

PR #1076, opened 2026-09-02
[Router][Bugfix] Log a warning when an engine scrape matches zero known metric names

### LINKED ISSUE #1074: [Bug][Router] Unrecognized metric names yield an all-zero EngineStats that looks healthy — load-aware routing silently degrades to random
### Describe the bug

`EngineStats.from_vllm_scrape()` recognises exactly six metric names, all `vllm:`-prefixed (`src/vllm_router/stats/engine_stats.py:63-86` on `main` / `fc00f98`). Every sample that does not match one of the six is silently discarded, and the function then **unconditionally constructs and returns an `EngineStats`** from its zero-initialised locals:

```python
num_running_reqs = 0
num_queuing_reqs = 0
...
for family in text_string_to_metric_families(vllm_scrape):
    for sample in family.samples:
        if sample.name == "vllm:num_requests_running":
        ...
return EngineStats(...)   # line 78 — always reached
```

So a backend whose `/metrics` is well-formed Prometheus text but uses a different name prefix produces a **fully-populated `EngineStats` with every field zero**. Nothing raises and nothing is logged.

That object is then treated as a successful scrape at `engine_stats.py:150`:

```python
engine_stats = self._scrape_one_endpoint(url)
if engine_stats:
    collected_engine_stats[url] = engine_stats
```

`EngineStats` is a plain dataclass with no `__bool__` or `__len__`, so an all-zero instance is truthy. The backend is stored, `get_health()` stays true, and the router reports a normal scrape.

### Consequence

Every consumer of engine stats sees `num_running_requests == num_queuing_requests == 0` for that backend, forever. Any load computation over the fleet is therefore identically zero across backends, so load-aware placement degrades to an arbitrary or random pick with no signal — while looking, from every metric and log line the router emits, like it is working. There is no error, no warning, no unhealthy endpoint, and no way to tell the difference between "all backends are idle" and "the parser understood nothing".

The specific consumer that made us find it was our own least-loaded fallback (see #1073), which computes `num_running_requests + num_queuing_requests` per endpoint; it was inert for a whole model lane and we could not tell from the outside. But the same silent-degradation applies to anything reading these fields, including the dashboard gauges and `gpu_cache_usage_perc`, which will read 0% on an engine at capacity.

### How we hit it

Two of our backends are **SGLang** rather than vLLM. SGLang's Prometheus exporter uses an `sglang:` prefix — `sglang:num_running_reqs`, `sglang:num_queue_reqs`, `sglang:token_usage`, `sglang:cache_hit_rate` — so all six comparisons miss on every sample and the result is the all-zero object described above.

That the router is vLLM-focused is entirely reasonable; the bug is not the missing SGLang support, it is that **a total parse miss is indistinguishable from an idle engine**. The same thing would happen to a vLLM version that renamed a metric, to a `/metrics` endpoint fronted by a proxy that strips or re-prefixes names, or to any other engine someone points the router at.

### Suggested fix

Either of these would have surfaced it immediately, and we have no preference between them:

1. **Log loudly.** Count how many samples matched; if zero names matched but the payload parsed into at least one metric family, `logger.warning` once per endpoint naming the URL and a few of the names that were actually present. Cheap, no behaviour change, and it turns a silent failure into an obvious one.
2. **Return `None`.** Treat "parsed, but recognised nothing" as a failed scrape, which `_scrape_one_endpoint`'s callers already handle (the endpoint is simply absent from `engine_stats`), letting load-aware consumers fall back to whatever they do when stats are missing rather than acting on fabricated zeros. This is the stronger fix but is a behaviour change, so it may want to be opt-in.

Option 1 alone would be enough to make this class of problem findable.

### Also happy to contribute

We carry a four-line name mapping for the SGLang equivalents (`sglang:num_running_reqs` → `num_running_requests`, `sglang:num_queue_reqs` → `num_queuing_requests`, `sglang:token_usage` → `gpu_cache_usage_perc`, `sglang:cache_hit_rate` → `gpu_prefix_cache_hit_rate`, all verified as gauges on a live SGLang `/metrics`, units matching vLLM's 0–1 fraction convention). It cannot regress vLLM backends, since a vLLM engine never emits an `sglang:` name and vice versa. We deliberately did **not** map the two prefix-cache *counters*: SGLang's nearest counters (`sglang:cached_tokens_total`, `sglang:prefill_effective_tokens_total`) count tokens, not cache queries, so mapping them would put a wrong-unit number behind a right-sounding field name.

If a multi-engine metric mapping is in scope for this component we'll open that as a PR; if it is explicitly out of scope, then the detection/logging fix above is the part that matters and we'll send that instead. Say which and we'll do it.

### Environment

Router `main`, static service discovery, two vLLM backends and two SGLang backends behind one router.

pujitha24:
[Router][Bugfix] Log a warning when an engine scrape matches zero known metric names

FIX #1074 (*link existing issues this PR will resolve*)

## Motivation

`EngineStats.from_vllm_scrape()` recognizes exactly six `vllm:`-prefixed
metric names. If a `/metrics` payload parses as valid Prometheus text but
none of its sample names match one of those six (e.g. an SGLang backend,
which uses an `sglang:` prefix, or any other non-vLLM/renamed exporter),
the function still unconditionally returns a fully-populated, all-zero
`EngineStats`. That object is truthy, so it's stored as a successful
scrape, `get_health()` stays true, and load-aware routing silently
degrades to an arbitrary/random pick for that backend — with no error,
warning, or any other signal that the parser understood nothing.

The issue offered two fixes with no preference between them: (1) log a
warning when a scrape parses but matches zero known names, described as
"cheap, no behaviour change" and sufficient on its own "to make this
class of problem findable", or (2) return `None` to treat it as a failed
scrape, described by the issue author as the stronger fix but a
behavior change that "may want to be opt-in". This PR implements option
1 only, to keep the change minimal and non-controversial.

## Approach

In `EngineStats.from_vllm_scrape()`, track how many samples matched a
known `vllm:` metric name and which sample names were actually seen.
After parsing, if at least one sample was seen but none matched, log
`logger.warning` naming the endpoint URL and up to 5 of the metric names
that were present, so the mismatch is now visible in the router's logs
instead of silently masquerading as an idle-but-healthy backend. A
scrape with at least one matched name (a partial match) is treated as a
partial success and does not warn.

`from_vllm_scrape` now takes an optional `url` parameter (used only for
the log message); its return value and default behavior are otherwise
unchanged — it still always returns an `EngineStats`, never `None`. The
only caller, `_scrape_one_endpoint`, was updated to pass the URL
through.

## Validation

Ran the repo's `pre-commit` hooks (black, isort, ruff, codespell,
trailing-whitespace, end-of-file-fixer) against the two changed files —
all passed.

Added `src/tests/test_engine_stats.py` with 4 tests covering: normal
vLLM metrics parse correctly and emit no warning; an SGLang-style scrape
(zero matching names) parses to an all-zero `EngineStats` *and* emits
the new warning naming the endpoint and the unrecognized metric names;
an empty payload does not warn; a mixed payload with at least one
matching name does not warn (partial match is not treated as total
failure). Ran:

```
python -m pytest src/tests/test_engine_stats.py -v
```

All 4 passed. I also confirmed the tests are meaningful by stashing the
fix and re-running: all 4 fail against the pre-fix code (the old
function signature doesn't accept `url`, and the warning is absent).

Note: I could not run `uv sync --all-extras` in this environment — the
`test` extra pulls in `vllm==0.13.0`, which requires GPU-only wheels
(`cupy-cuda12x`, `nvidia-cudnn-frontend`) unavailable on macOS/arm64. I
instead installed only this package's base dependencies plus
pytest/pytest-asyncio into a venv to run the affected test file; the
changed code (`EngineStats`/`from_vllm_scrape`) has no dependency on
`vllm` or GPU packages, so this should be equivalent for this change. I
was not able to run the full Router E2E / helm chart CI workflows
locally.

---

---

<details>

<ul>
</ul>

<h3>Code Quality</h3>

<ul>
</ul>

, Shaoting-Feng or ApostaC.
</details>


pujitha24:
Good catches, fixed both: `url` is now `Optional[str]` instead of `str = None`, and the warning now joins the metric names with `", "` instead of logging the raw list repr. Ran the test suite locally (`pytest src/tests/test_engine_stats.py`) and pre-commit — all still pass.

pujitha24:
This has been green and addressed the earlier review feedback for a few days now — let me know if anything else would help move it along, @ApostaC.

src/vllm_router/stats/engine_stats.py
```diff
@@ -40,12 +40,14 @@ class EngineStats:
...
 
     @staticmethod
-    def from_vllm_scrape(vllm_scrape: str):
+    def from_vllm_scrape(vllm_scrape: str, url: Optional[str] = None):
         """
         Parse the vllm scrape string and return a EngineStats object
 
         Args:
             vllm_scrape (str): The vllm scrape string
+            url (str): The URL the scrape was fetched from, used only for
```
Shaoting-Feng: The docstring still says `url (str)`, but the type annotation has been changed to `Optional[str]`.

pujitha24:
Good catch, fixed — the docstring now says `url (Optional[str])` to match the signature.

Shaoting-Feng (approved):
LGTM

PR #1088, opened 2026-09-13
[Bugfix][Router] Soften fastapi pin to allow patched starlette

### LINKED ISSUE #1078: bug: exact fastapi==0.128.0 pin blocks security patches in transitive starlette
### Describe the bug

`pyproject.toml` pins FastAPI exactly:

```toml
"fastapi==0.128.0",
```

FastAPI 0.128.0 declares `starlette<0.51.0,>=0.40.0`. That upper bound makes it impossible for any downstream consumer to take a patched Starlette when Starlette ships a security fix above it.

Concretely, today: [GHSA-jp82-jpqv-5vv3](https://github.com/advisories/GHSA-jp82-jpqv-5vv3) (CVE-2026-54282) affects `starlette < 1.3.0`. Because the exact FastAPI pin caps Starlette below `0.51.0`, a `vllm-router` install cannot resolve to a fixed Starlette at all — the advisory stays open for every consumer until this pin moves.

This is not specific to one CVE. An exact pin on a framework that itself constrains a transitive dependency means every future FastAPI or Starlette advisory is blocked on a `production-stack` release, even when the fix is a patch-level bump the router would not otherwise notice.

### To Reproduce

```bash
uv pip compile --python-version 3.12 - <<'REQ'
vllm-router @ git+https://github.com/vllm-project/production-stack@main
REQ
```

Resolves `fastapi==0.128.0` and `starlette==0.50.0` (vulnerable).

Adding an override to force a patched Starlette fails:

```
Because fastapi==0.128.0 depends on starlette<0.51.0,>=0.40.0 and you require
starlette>=1.3.0, we can conclude that your requirements are unsatisfiable.
```

The only way through is to override FastAPI itself — which resolves cleanly to `fastapi==0.141.1` + `starlette==1.6.0`, but silently moves the router 13 FastAPI minors away from the version upstream tests.

### Expected behavior

A range rather than an exact pin, so consumers can take security patches without overriding the framework — e.g. `fastapi>=0.128.0,<1`, or a floor that admits a Starlette above the advisory.

### Additional context

We're a downstream consumer building a slim router image, and we hit this while clearing low-severity dependency advisories for a compliance control. Our own exposure to this particular CVE looks negligible — it poisons `request.url.hostname`/`netloc`, and we found no reads of those attributes anywhere in `src/` (0 hits across 92 files; the only `request.url` uses are `.path` as a metric label and `str(request.url)` as a span attribute) — so we are not asking for an urgent fix. The reason for filing is the structural one: the exact pin is what forces every consumer into an override.

**A judgment that's yours, not ours:** we don't know whether the exact pin is deliberate — whether the router depends on 0.128-specific FastAPI behaviour, or whether it's pinned for reproducibility and a range would be fine. If it's the latter, we're happy to open a PR relaxing it, with the DCO sign-off. If it's the former, a comment recording *why* would be just as useful to downstream consumers, since it would tell us to stop trying.

---

*Disclosure: this issue was authored by Claude (Anthropic's AI assistant, via Claude Code) on behalf of @MichelFloyd, who reviewed and takes responsibility for it.*

dundysm:
## Summary
exact `fastapi==0.128.0` declares `starlette<0.51.0`, so consumers cannot resolve patched starlette (e.g. GHSA-jp82-jpqv-5vv3 / CVE-2026-54282, fixed in `starlette>=1.3.0`) without overriding fastapi itself.

soften the pin to `fastapi>=0.128.0,<1` in both `pyproject.toml` and `src/vllm_router/requirements.txt` so a patched starlette can resolve. fastapi >=0.133.0 no longer upper-bounds starlette; default resolution picks a current fastapi + starlette>=1.3.

Fixes #1078

## Test plan
- [x] `uv pip compile` with `fastapi==0.128.0` + `starlette>=1.3.0` fails (baseline)
- [x] `uv pip compile` with `fastapi>=0.128.0,<1` + `starlette>=1.3.0` resolves to `fastapi==0.141.1` + `starlette==1.6.0`
- [x] local package + `starlette>=1.3.0` compiles cleanly
- [x] `uv pip install -e ".[test]"` then `pytest src/tests/test_*.py` -> 173 passed (fastapi 0.141.1, starlette 1.6.0)

## notes
minimal range change only; left `uv.lock` alone (already lagged pyproject at 0.115.8). cc @ruizhang0101 for a light look if useful.


dundysm:
@ruizhang0101 friendly ping when you have a moment; softens the exact fastapi==0.128.0 pin so patched starlette can resolve (fixes #1078). router unit tests 173 passed on fastapi 0.141.1 / starlette 1.6.0.

Shaoting-Feng (approved):
LGTM

dundysm:
@Shaoting-Feng thanks for the lgtm; this one's approved + checks clean whenever you're ready to squash-merge.

PR #1089, opened 2026-09-13
[Bugfix][Router] Use least-inflight for PrefixAware below-threshold fallback

### LINKED ISSUE #1073: [Bug][Router] PrefixAwareRouter's below-threshold fallback uses stale QPS and starves one backend under session affinity
### Describe the bug

`PrefixAwareRouter`'s below-threshold branch falls back to `_qps_routing`. On `main` (`fc00f98`), `src/vllm_router/routers/routing_logic.py:841`:

```python
if match_length < self.prefix_min_match_length:
    # Fall back to QPS routing, but still record the prompt in the
    # trie. ...
    selected_endpoint = self._qps_routing(endpoints, request_stats)
    if selected_endpoint is not None:
        await self.hashtrie.insert(prompt, selected_endpoint)
    return selected_endpoint
```

`_qps_routing` picks the endpoint with the lowest `request_stats[url].qps`, which is a **rate averaged over the request-stats window** — a historical quantity, not a measure of what the engine is doing right now. Under session-shaped traffic that produces a stable positive feedback loop:

1. Backend B is mid-session and holds several long-running requests. Its QPS — completed requests per second over the window — is *low*, because long requests complete rarely.
2. A new session's first request matches below `prefix_min_match_length`, so it takes this branch, and QPS picks B: the busiest backend looks like the least busy one.
3. The prompt is then inserted into the trie against B (correctly — that is #990's fix), so every subsequent turn of that session has prefix affinity to B and never reaches this branch again.

Step 3 is what makes this more than transient noise: the sub-threshold decision is not re-evaluated per request, it is *latched* for the lifetime of the session. A bad first pick becomes a permanent one.

### Measured behaviour

Two vLLM backends, equal capacity, `prefixaware` routing, eight concurrent agent sessions opened in a burst. Result: **4 running + 4 queued on one backend while the other sat idle** — half the fleet's capacity unused and half the sessions queueing, with no error, no unhealthy backend, and nothing in the router logs to indicate it.

### What we run instead

Since 2026-08-18 we have run a shadow-patched `routing_logic.py` in production with the sub-threshold branch changed to least-in-flight, computed from engine stats rather than request stats:

```python
loads = {}
for info in endpoints:
    stats = engine_stats.get(info.url)
    if stats is None:
        return None  # stats not scraped yet; caller falls back to QPS
    loads[info.url] = stats.num_running_requests + stats.num_queuing_requests
min_load = min(loads.values())
near_tied = [u for u, l in loads.items() if l <= min_load + 1]
return random.choice(near_tied)
```

wired as:

```python
selected_endpoint = self._least_inflight_routing(
    endpoints, engine_stats
) or self._qps_routing(endpoints, request_stats)
```

Three properties that matter:

- **`num_running_requests + num_queuing_requests` is the actual instantaneous saturation** of the engine, which is the quantity this branch wants. `engine_stats` is already passed into every `route_request`; no new collection path.
- **Random choice among backends within +1 of the minimum**, rather than strict argmin. Engine stats refresh on the scrape interval, so a burst arriving inside one window would otherwise see identical loads and dogpile whichever URL sorts first — exactly the case that produced the 4+4 split. The jitter is what fixes the burst; strict least-loaded alone does not.
- **Falls back to `_qps_routing` when engine stats are absent** (backend not yet scraped, scrape failing), so behaviour on a cold router is unchanged, and #990's trie insert is preserved as-is.

This has been running for two weeks on the traffic shape described above with the imbalance gone.

### Related prior art

- #990 (merged) fixed the *other* bug on this same branch — the trie never being seeded — and our patch keeps that insert unchanged.
- #1035 (merged) addresses the analogous cache-affinity-vs-load tension for the `kvaware` family via `loadaware` scoring. `PrefixAwareRouter` has no equivalent: it still hands the sub-threshold case to QPS.

### Question before we send a PR

Would you prefer this as the new default for the sub-threshold branch (QPS retained only as the no-stats fallback, as above), or behind a flag — e.g. `--prefixaware-fallback={qps,leastinflight}` with `qps` as default? We have the change in hand and are happy to open it either way; tell us which shape you'd merge and we'll match it, with unit tests in the `src/tests/test_prefixaware_router.py` style.

### Environment

Router in front of two API-key-protected vLLM backends, static service discovery, `prefixaware` routing logic, `--engine-stats-interval 10`.

dundysm:
## Summary
Fixes #1073.

`PrefixAwareRouter`'s below-threshold branch previously fell back to `_qps_routing` (windowed QPS) and then inserted the choice into the hashtrie. Under session-shaped traffic that creates a positive feedback loop: a backend mid long-running requests has *low* completed QPS while actually saturated; QPS picks it; the trie latches that pick for the session; capacity sits idle on the other backend.

## Policy choice
**Least-inflight as the new default** for the sub-threshold branch; QPS retained only when engine stats are unavailable (cold scrape / partial scrape). No feature flag.

Why default rather than `--prefixaware-fallback={qps,leastinflight}`:
- the QPS path is the bug under the traffic shape the router is meant to serve with `prefix_min_match_length > 0`
- `engine_stats` is already passed into every `route_request`; no new collection path
- the reporter has run this shape in production for ~2 weeks with the imbalance gone
- a flag would leave the starvation path as the shipped default

Behaviour:
1. compute load as `num_running_requests + num_queuing_requests`
2. choose uniformly at random among backends within `+1` of the minimum (avoids dogpiling inside one scrape window)
3. if any endpoint lacks engine stats, fall back to `_qps_routing` (unchanged cold-router behaviour)
4. keep the #990 trie insert on the selected endpoint so later turns can pin

Sibling #1074 is a different bug (unrecognized metrics -> all-zero EngineStats) and is not addressed here.

## Test plan
- [x] `pytest src/tests/test_prefixaware_router.py` (8 passed)
- [x] related router suite: roundrobin / session / priority / loadaware / kvaware (66 passed)
- [x] new coverage:
  - stale low-QPS / high in-flight starvation case prefers the idle backend
  - incomplete engine_stats falls back to QPS
  - near-tied in-flight spreads across the +1 band (no dogpile)
  - real HashTrie: least-inflight seed + follow-up pin away from the busy backend
  - existing #990 seeding / above-threshold / default-threshold-0 cases still pass

## Risks
- when engine scrape is healthy, sub-threshold routing no longer follows QPS; operators who relied on QPS specifically for that branch will see different first-picks (intended)
- if scrape is stale but present, least-inflight can still be wrong until the next scrape; +1 jitter mitigates burst dogpiling but not a stuck scrape
- interaction with #1074: all-zero EngineStats that look healthy would make every backend appear idle and reduce to near-tie random among all; that is still better than QPS starvation, but a separate fix
- `prefix_min_match_length == 0` path unchanged (no below-threshold fallback)


dundysm:
@ruizhang0101 friendly ping when you have a moment; PrefixAware below-threshold fallback now uses least-inflight instead of stale windowed qps under session-shaped traffic (fixes #1073). router unit tests green including a starvation repro.

PR #1092, opened 2026-09-16
[Bugfix][Router] Reconcile the engine list when the K8s watch reconnects

### LINKED ISSUE #1091: bug: router keeps a deleted pod as a healthy backend when a DELETED event is missed during a watch reconnect
### Your current environment

vLLM Production Stack router (`lmcache/lmstack-router`, `main` @ `ebbb862`), Kubernetes service discovery in `pod-ip` mode, on a scale-to-zero GPU cluster (KEDA scales each vLLM worker 0↔1).

### Describe the bug

`K8sPodIPServiceDiscovery._watch_engines()` treats the watch stream as its only source of engine removals. When the stream drops, the loop reconnects but never re-lists, so **any `DELETED` event that occurs while the connection is down is lost permanently** and the engine stays in `available_engines` forever.

The router then round-robins onto a pod IP that no longer exists. Requests hang until they fail with `ClientConnectorError` (~130 s in our case), while `/v1/models` and `/health` stay green because both are served from the router's own state — so every health check and uptime probe reports the service as healthy.

This is distinct from #1008 / #1012, which fixed the eviction case (`MODIFIED` with `podIP=None`). We run an image that already contains that fix; the pod here is deleted normally, the event is simply never delivered.

In production this cost us **22 hours of silently failing inference** (2026-09-15 16:20 → 2026-09-16 14:07), with a `K8s watcher error: Response ended prematurely` every ~10-15 minutes:

```
[2026-09-16 13:58:06] INFO:  Scraping metrics from 1 serving engine(s)
[2026-09-16 13:58:21] ERROR: Failed to scrape metrics from http://10.2.1.14:8000: Connection timed out
```

while `kubectl get endpoints -n vllm` showed no backend at all, and:

```
vllm:healthy_pods_total{server="http://10.2.1.14:8000"} 1.0
vllm:num_requests_running{server="http://10.2.1.14:8000"} 124.0      # never drains
vllm:request_errors_total{error_type="ClientConnectorError", ...} 124.0
```

`num_requests_running` is never decremented on error, so the ghost also skews least-loaded routing decisions.

### Reproduction

Deterministic, with a NetworkPolicy that blocks the router's egress to the API server (the real control-plane IP, not the `kubernetes.default` ClusterIP — kube-proxy DNATs before the policy applies):

1. Let the router discover a worker (`vllm:healthy_pods_total` shows its pod IP).
2. Apply the policy and **wait for the watch to drop on its own** — an established connection survives the policy, so the router is only truly blind after its next reconnect attempt fails.
3. Delete the worker pod during that window.
4. Remove the policy.

The router reconnects and keeps serving the deleted pod indefinitely: `/v1/models` still advertises the model while no pod exists, and a real request times out.

### Suggested fix

List the pods on every watch (re)connection and drop the engines that no longer exist — the LIST+WATCH pattern every Kubernetes controller uses. It only runs when the stream restarts, so there is no periodic cost in steady state. PR follows.

### Related

`K8sServiceNameServiceDiscovery` has the mirror-image problem, which makes `service-name` unusable with scale-to-zero: it watches Services, but a pod becoming Ready does not modify the Service (the Endpoints object changes), so no event fires and `_check_service_ready()` is never re-evaluated. We measured a worker `1/1 Ready` for 6+ minutes with the router still at `0 serving engine(s)` and `/v1/models` empty. Happy to open a separate issue if useful.

jStrider:
FIX #1091

`K8sPodIPServiceDiscovery._watch_engines()` treats the watch stream as its only source of engine removals. When the stream drops, the loop reconnects but never re-lists, so a `DELETED` event delivered while the connection is down is lost permanently: the engine stays in `available_engines` and keeps receiving traffic.

Requests then round-robin onto a pod IP that no longer exists and hang until `ClientConnectorError`, while `/v1/models` and `/health` stay green — both are served from the router's own state, so health checks never notice. `vllm:num_requests_running` for that server climbs and never drains, which also skews least-loaded routing.

This is distinct from #1008 / #1012, which fixed the eviction case (`MODIFIED` with `podIP=None`). Here the pod is deleted normally; the event is simply never delivered.

## The change

`_reconcile_engines()` lists the pods and drops the engines whose pod is gone. It is called at the top of the `while self.running` loop, so it runs before each stream is opened — that is, exactly when events may have been missed, and only once at startup in steady state. No new thread, no new flag, no periodic cost.

Removal happens under `available_engines_lock` rather than through `_delete_engine()`, which takes the same lock and would `del` without a guard — avoiding both a deadlock and a race with the watch thread.

## Testing

Two unit tests in the existing `test_k8s_pod_ip_service_discovery.py`: an engine whose pod is gone is dropped, and live engines are never dropped. Both fail if the detection is neutralised (verified by mutating `stale` to an empty set).

Validated end to end on a scale-to-zero GPU cluster (KEDA, workers 0↔1), same protocol both times — block the router's egress to the API server, wait for the watch to drop on its own, delete the worker while blind, restore the network:

| | without the patch | with the patch |
|---|---|---|
| ghost present on restore | yes | yes |
| after 30 s | still served | `no longer exists but was still registered: dropping it` |
| after 3 min | `/v1/models` advertises the model, real request times out at 45 s | engine list empty |

Note for anyone reproducing this: the NetworkPolicy must exclude the real control-plane IP, not the `kubernetes.default` ClusterIP (kube-proxy DNATs before the policy applies), and an already-established watch survives the policy — the router is only truly blind once its next reconnect fails.


jStrider:
Thanks — both points were right, and acting on them found a third problem. Pushed a second commit.

**Race between the list and the watch.** Fixed as suggested: `_reconcile_engines()` now returns `pods.metadata.resource_version` and the watch starts from it. This also removes the redundant `/v1/models` request to every pod on each reconnect, which was real: `timeout_seconds` is always passed, so `disable_retries` is `True` in `watch.py` and every `stream()` call restarted without a `resource_version` — a full list plus an `ADDED` per pod, every time.

One consequence worth stating explicitly: with a `resource_version`, the stream no longer replays `ADDED`, so the reconciliation is now the **only** discovery path, not just a cleanup. It therefore populates as well as removes, and a ready pod already registered at the same URL, model label and sleep label is skipped so the steady state still costs a single list. I verified on a live cluster that discovery still works — a cold-started worker is picked up and served normally.

**`K8sServiceNameServiceDiscovery`.** Same treatment, it had the identical hole.

**Third problem, found while doing the above.** Because the reconciliation is now the discovery path, a single failing object was enough to break everything: a `Service` matched by the selector but with no `Endpoints` object (headless, `ExternalName`, or just created) makes `_check_service_ready` raise, the exception escapes to `_watch_engines`, and the loop retries at 2 Hz with **no engine ever discovered**. Each object is now reconciled in isolation.

**Known limitation**, not addressed here: the reconciliation is sequential, and an unknown pod that is k8s-ready but HTTP-unreachable costs `health_check_timeout_seconds`. With enough such pods the pass could outlive the API server's `resource_version` retention, yielding a `410` and another list — self-healing, but it would stay list-only without ever streaming. Worth a follow-up if anyone runs into it.

**Testing.** 13 unit tests covering startup discovery, the skip, readiness and cleared-IP transitions missed during a disconnect, sleep-label changes, `resourceVersion` propagation, per-object isolation, and the service-name equivalents. Each one was checked against a mutation that neutralises the behaviour it asserts.

End to end on a scale-to-zero GPU cluster, same protocol as before — block the router's egress to the API server, wait for the watch to drop, delete the worker while blind, restore:

```
WARNING: Serving engine ...-b47s6 no longer exists but was still registered: dropping it
engines still listed: NONE
```

Against the unpatched image, the same sequence leaves `/v1/models` advertising the model with no pod behind it and a real request timing out at 45 s.

src/vllm_router/service_discovery.py
```diff
@@ -668,55 +668,124 @@ def _get_model_label(self, pod) -> Optional[str]:
...
+        ) and not self._is_pod_terminating(pod)
+        if not is_ready:
+            return False
+
+        labels = pod.metadata.labels or {}
+        return (
+            known.url == f"http://{pod.status.pod_ip}:{self.port}"
+            and known.model_label == self._get_model_label(pod)
+            and known.sleep == (labels.get("sleeping") == "true")
+        )
```
Shaoting-Feng, line 734: Should we also compare the pod's `lora-modified` annotation (store it on `EndpointInfo`)?

jStrider: Yes — good catch, it was a real hole. Done in 7adf99a.

`lora-modified` is the annotation the operator stamps on the pod (`triggerPodEvent`, `loraadapter_controller.go`) for the sole purpose of forcing the router to re-read `/v1/models`. It is the only pod-side signal that the adapter list moved: when an adapter is loaded or unloaded, the URL, the readiness, the `model` label and the `sleeping` label are all identical before and after. So `_is_unchanged` returned `True` and, on reconnect, a `lora-modified` event lost during the disconnect was swallowed — the router kept serving a stale adapter list until the next real event on that pod. Same class of bug as the one this PR fixes, applied to adapters.

`EndpointInfo` now carries `lora_modified`, `_handle_pod` reads it through `_get_lora_modified()` and passes it down to `_add_engine`, and `_is_unchanged` compares it. New test `test_reconcile_rehandles_pod_whose_lora_annotation_changed`, checked against a mutation removing the comparison.

`K8sServiceNameServiceDiscovery` is deliberately untouched on this point: the operator annotates pods, not Services, so there is no equivalent signal to compare there.

The same commit also applies `black` to the two previous ones, which is what CI was failing on.

PR #217, opened 2025-03-03
refactor: standard fastapi project structure for better main…

BrianPark314:
…tainability

FILL IN THE PR DESCRIPTION HERE

This is a suggestion for restructuring the project into more maintainable shape. Changes include:

apply mvc pattern
isolate the entry point
separate routers
removed duplicated code (in batch router)
fix typo
Hope this gets reviewed before 0.1.0 release!

**BEFORE SUBMITTING, PLEASE READ THE CHECKLIST BELOW AND FILL IN THE DESCRIPTION ABOVE**

---

- [ ] Make sure the code changes pass the [pre-commit](../CONTRIBUTING.md) checks.
- [ ] Sign-off your commit by using <code>-s</code> when doing <code>git commit</code>
- [ ] Try to classify PRs for easy understanding of the type of changes, such as `[Bugfix]`, `[Feat]`, and `[CI]`.

<details>

<summary><b> Detailed Checklist (Click to Expand) </b></summary>

<p>Thank you for your contribution to production-stack! Before submitting the pull request, please ensure the PR meets the following criteria. This helps us maintain the code quality and improve the efficiency of the review process.</p>

<h3>PR Title and Classification</h3>
<p>Please try to classify PRs for easy understanding of the type of changes. The PR title is prefixed appropriately to indicate the type of change. Please use one of the following:</p>
<ul>
    <li><code>[Bugfix]</code> for bug fixes.</li>
    <li><code>[CI/Build]</code> for build or continuous integration improvements.</li>
    <li><code>[Doc]</code> for documentation fixes and improvements.</li>
    <li><code>[Feat]</code> for new features in the cluster (e.g., autoscaling, disaggregated prefill, etc.).</li>
    <li><code>[Router]</code> for changes to the <code>vllm_router</code> (e.g., routing algorithm, router observability, etc.).</li>
    <li><code>[Misc]</code> for PRs that do not fit the above categories. Please use this sparingly.</li>
</ul>
<p><strong>Note:</strong> If the PR spans more than one category, please include all relevant prefixes.</p>

<h3>Code Quality</h3>

<p>The PR need to meet the following code quality standards:</p>

<ul>
    <li>Pass all linter checks. Please use <code>pre-commit</code> to format your code. See <code>README.md</code> for installation.</li>
    <li>The code need to be well-documented to ensure future contributors can easily understand the code.</li>
    <li> Please include sufficient tests to ensure the change is stay correct and robust. This includes both unit tests and integration tests.</li>
</ul>

[...]

PR #1093, opened 2026-09-17
[Bugfix][Router] follow the batch module through the rename that moved it

chrikrah:
#217 renamed `src/vllm_router/batch/` to `src/vllm_router/services/batch_service/` in `95efecc`, on 2025-03-04. `local_processor.py` moved with it and kept importing `BatchInfo` and `BatchStatus` from the old path, at line 24 and again inside `retrieve_batch` at line 127. The module it wants is now its own sibling.

```
$ gh api 'repos/vllm-project/production-stack/commits/95efecc' --jq '.files[] | select(.filename|test("batch")) | "\(.status) \(.filename) <- \(.previous_filename)"'
renamed src/vllm_router/services/batch_service/batch.py <- src/vllm_router/batch/batch.py
renamed src/vllm_router/services/batch_service/local_processor.py <- src/vllm_router/batch/local_processor.py
```

`app.py:254` calls `initialize_batch_processor` whenever a user passes `--enable-batch-api`. That function imports `LocalBatchProcessor`, so the router aborts at startup:

```
$ python -c "from vllm_router.services.batch_service import initialize_batch_processor; initialize_batch_processor('local', '/tmp/store', None)"
  File ".../services/batch_service/local_processor.py", line 24, in <module>
    from vllm_router.batch.batch import BatchInfo, BatchStatus
ModuleNotFoundError: No module named 'vllm_router.batch'
```

#784 fixed the same orphaned import one layer out, in `services/batch_service/__init__.py`, and stopped there.

Line 22 of the same file imports `aiosqlite`. Neither `pyproject.toml` nor `src/vllm_router/requirements.txt` declares it, so a stock install fails one line earlier. `router-e2e-test.yml:47` installs from `pyproject.toml` and line 80 collects `src/tests/test_*.py`, which now includes the new file, so this branch declares it in both places.

Two regression tests cover the path `app.py` takes. One builds the processor through `initialize_batch_processor`. The other creates a batch and reads it back. Run against the pinned `aiosqlite==0.21.0`:

```
$ python -m pytest src/tests/test_batch_service.py -q
2 passed in 0.04s
```

Revert the import change, keep the tests, and both fail:

```
2 failed in 0.04s
FAILED src/tests/test_batch_service.py::test_initialize_local_batch_processor
FAILED src/tests/test_batch_service.py::test_create_and_retrieve_batch
```

Correcting only the module-level import leaves the second one failing, which is what the line 127 edit is for.

No open issue covers this, so there is nothing to link. #766 reported the sibling defect and #784 closed it.

---

One limit: no live vLLM backend served these endpoints, so this covers startup and the SQLite round trip only. `cancel_batch` also writes the literal `"cancelled"`, which `BatchStatus` has no member for, so `retrieve_batch` raises `ValueError` afterwards. That is a separate defect and stays out of this change.


chrikrah:
Both taken, in `78962ad`.

`tmp_path` replaces the hardcoded directory and its cleanup fixture. The hardcoded form came from the sibling `test_file_storage.py`, which does the same thing with `/tmp/test_vllm_files`, so this file no longer matches that one. `tmp_path` is the better pattern and it does not collide when CI runs tests concurrently.

The local `BatchStatus` import inside `retrieve_batch` existed to work around the broken module-level one. With that corrected it is dead weight, so it is gone rather than corrected in place.

Replacing the whole file with the `origin/main` copy still gives `ModuleNotFoundError` on both tests, so the module-level import is what they pin.

Shaoting-Feng (approved):
LGTM

PR #1094, opened 2026-09-17
[Bugfix][Router] carry the three label fields through DynamicRouterConfig.from_args

chrikrah:
`DynamicRouterConfig` declares `static_model_labels` (line 58), `prefill_model_labels` (line 63) and `decode_model_labels` (line 64). `parser.py` defines `--static-model-labels` (line 176), `--prefill-model-labels` (line 432) and `--decode-model-labels` (line 439). `reconfigure_service_discovery` reads all three, at lines 185, 190 and 193. `from_args` assigns none of them, so all three stay `None`.

On `ebbb862`, with `parse_args()` over a static plus disaggregated command line, then `from_args`:

```
args.static_model_labels   = 'prefill,decode'
cfg.static_model_labels    = None
args.prefill_model_labels  = 'prefill'
cfg.prefill_model_labels   = None
args.decode_model_labels   = 'decode'
cfg.decode_model_labels    = None
```

`app.py:260` builds this object as `init_config` and hands it to the watcher, which keeps it as `current_config` until the config file first changes. `main_router.py:242` serialises `current_config` into `/health`, so `/health` reports three fields as null that the operator set. `parser.py:540` merges a dynamic config file into `args` as defaults before `from_args` runs, so the drop applies whether the values arrived on the command line or in that file.

The fix is three added lines, beside the fields already assigned from the same `args`.

Two regression tests. The first asserts the three values survive `from_args`. The second asks, for every field the dataclass and the parser share, whether it moved off its dataclass default.

Asking about the default rather than about equality with the parsed argument is deliberate, and two earlier versions of that test proved why. Comparing for equality passed for eight of twenty-one fields, because argparse and the dataclass are written to share their defaults, so both sides read the same even with the assignment deleted. Comparing for truthiness then still missed `static_backend_health_check_interval` and `priority_header`, whose defaults are themselves truthy.

Deleting each of nine assignments in turn now fails the test in every case:

```
drop priority_header                      -> 1 failed, 1 passed
drop static_backend_health_check_interval -> 1 failed, 1 passed
drop callbacks                            -> 1 failed, 1 passed
drop priority_default                     -> 1 failed, 1 passed
drop static_model_labels                  -> 2 failed
drop prefill_model_labels                 -> 2 failed
drop decode_model_labels                  -> 2 failed
drop session_key                          -> 1 failed, 1 passed
drop static_aliases                       -> 1 failed, 1 passed
```

```
$ python -m pytest src/tests/ -q
233 passed in 14.42s
```

The default comparison also tolerates a transformation such as `parse_comma_separated_args` moving into `from_args` later: a transformed value still differs from the declared default.

No open issue covers this, so there is nothing to link.

---

Two things stay out of this change. `k8s_service_discovery_type` and `k8s_watcher_timeout_seconds` never reach the dataclass at all, so `from_args` cannot drop them; `app.py:228` and `:235` pass both at startup while the k8s branch of `reconfigure_service_discovery` passes neither, which is a larger defect than this one and belongs on its own. And no live deployment exercised the reconfigure path, so this covers `from_args` and the `/health` payload it feeds.


chrikrah:
Correct, and it was the third form of that blindness in this one test.

`--priority-field` is now `request_priority`. An earlier version compared each shared field for equality with the parsed argument, which passed for eight of twenty-one fields because argparse and the dataclass are written to share their defaults. The version before this one compared for truthiness, which still missed `static_backend_health_check_interval` and `priority_header`, whose defaults are themselves truthy. Comparing against the dataclass default fixed those two and left this one, because `ARGV` happened to pass the default value.

The test now asserts up front that no shared field sits at its dataclass default, so the next field added without a distinct value fails here rather than passing quietly.

Deleting each of the eighteen assignments in `from_args` in turn now fails the test in every case. `f275358`.

src/vllm_router/dynamic_config.py
Shaoting-Feng, line 94: Optional simplification:

```python
@staticmethod
def from_args(args) -> "DynamicRouterConfig":
    return DynamicRouterConfig(
        **{f.name: getattr(args, f.name) for f in dataclasses.fields(DynamicRouterConfig)}
    )
```

What do you think?

PR #1021, opened 2026-07-28
[Router][Bugfix] Handle all router types correctly in proxy_multipart_request

shernshiou:
## Description

`proxy_multipart_request` (used by `/v1/audio/transcriptions` and `/v1/images/edits`) was calling `router.route_request()` synchronously and without the required `request_json` argument, causing a `TypeError` when using `SessionRouter`, `KvawareRouter`, or `PrefixAwareRouter`.

## Changes

- Added the same `isinstance(router, (KvawareRouter, PrefixAwareRouter, SessionRouter))` guard already used in `route_general_request`
- For async routers: call with `await` and pass `{}` as `request_json` (no JSON body in multipart/form-data)
- For other routers (e.g. `RoundRobinRouter`): keep the original synchronous call unchanged

Fixes #1019

**BEFORE SUBMITTING, PLEASE READ THE CHECKLIST BELOW AND FILL IN THE DESCRIPTION ABOVE**

---

- [x] Make sure the code changes pass the [pre-commit](https://github.com/vllm-project/production-stack/blob/main/CONTRIBUTING.md) checks.
- [x] Sign-off your commit by using <code>-s</code> when doing <code>git commit</code>
- [ ] Try to classify PRs for easy understanding of the type of changes, such as `[Bugfix]`, `[Feat]`, and `[CI]`.

<details>

<summary><b> Detailed Checklist (Click to Expand) </b></summary>

<p>Thank you for your contribution to production-stack! Before submitting the pull request, please ensure the PR meets the following criteria. This helps us maintain the code quality and improve the efficiency of the review process.</p>

<h3>PR Title and Classification</h3>
<p>Please try to classify PRs for easy understanding of the type of changes. The PR title is prefixed appropriately to indicate the type of change. Please use one of the following:</p>
<ul>
    <li><code>[Bugfix]</code> for bug fixes.</li>
    <li><code>[CI/Build]</code> for build or continuous integration improvements.</li>
    <li><code>[Doc]</code> for documentation fixes and improvements.</li>
    <li><code>[Feat]</code> for new features in the cluster (e.g., autoscaling, disaggregated prefill, etc.).</li>
    <li><code>[Router]</code> for changes to the <code>vllm_router</code> (e.g., routing algorithm, router observability, etc.).</li>
    <li><code>[Misc]</code> for PRs that do not fit the above categories. Please use this sparingly.</li>
</ul>
<p><strong>Note:</strong> If the PR spans more than one category, please include all relevant prefixes.</p>

[...]

PR #1095, opened 2026-09-18
[Bugfix][Router] dispatch PriorityRouter in proxy_multipart_request

chrikrah:
`PriorityRouter.route_request` is `async` and takes `request_json` (`routing_logic.py:917`). `route_general_request` lists `PriorityRouter` in both of its `isinstance` tuples, at `request.py:583` and `:637`. `proxy_multipart_request` lists it in neither of its branches, at `:1290` and `:1306`. So it falls through to the synchronous `else` at `:1315`, which passes four positional arguments where the method takes five.

That call sits outside the `try` at `:1324`. The `TypeError` therefore escapes as a bare 500 on `/v1/audio/transcriptions`, `/v1/audio/translations` and `/v1/images/edits`, for every request, whenever the router runs with `--routing-logic priority`.

```
FAILED src/tests/test_multipart_proxy.py::test_audio_translation_routes_under_every_routing_logic[priority]
E  TypeError: PriorityRouter.route_request() missing 1 required positional argument: 'request_json'
   src/vllm_router/services/request_service/request.py:1315: TypeError
```

## A stopgap, not the fix

#1021 corrected the identical omission for `SessionRouter`. While that sat in review, #1022 wrote down what would happen next: the dispatch "will break silently whenever a new router is added". `PriorityRouter` merged in #996 on 2026-08-18, after #1022 drew its table. It reached the two `route_general_request` call sites and not this one.

#1020 proposed this dispatch before #1021 merged it. Its tuple names three routers, predates `PriorityRouter`, and now conflicts. #1023 answers #1022 by removing the `isinstance` dispatch. That makes this line unnecessary rather than wrong. Both #1022 and #1023 have sat idle since 2026-07-29, and `main` returns 500 today, so one line can land now and #1023 deletes it later.

## Test

`test_multipart_proxy.py` hard-coded `ROUND_ROBIN`, the one routing logic the `else` branch handles correctly. Its `router_client` helper now takes the routing logic and defaults to `ROUND_ROBIN`, so the existing four tests do not change. One new test runs the translation path under both.

```
$ python -m pytest src/tests/test_multipart_proxy.py -q
6 passed in 7.16s

$ python -m pytest src/tests/ -q
233 passed in 15.00s
```

Revert the production line, keep the test, and the priority case fails with that `TypeError` while the round-robin case passes.

## Limits

The same dispatch line covers the two image and transcription endpoints, and no new test covers them, because the existing harness stands up only the audio path. No live vLLM backend ran. The backend here is the `aiohttp` test server the file already uses.

---


Shaoting-Feng (commented):
Thanks for your PR. Would you be interested in doing the root-cause fix as a follow-up PR? [vllm-project/production-stack#1023](https://github.com/vllm-project/production-stack/pull/1023) has been conflicting since 07/29/26 and the author hasn't responded to the rebase request, so it'd be good to have someone pick it up.

PR #937, opened 2026-04-29
[Feat][Router] Support multiple API keys for multi-tenant authentication (#759)

prashansapkota:
## Summary

Closes #759

Previously, `vllmApiKey` only accepted a single string value. Because vLLM's `--api-key` flag natively supports a comma-separated list of keys, users tried passing `"key1,key2,key3"` or a YAML list — but the Helm chart silently discarded everything after the first key, and the router had no enforcement layer at all. This blocked multi-tenant deployments where different teams need independent credentials to share the same vLLM instance.

This PR fixes the full stack. From Helm values through to the router's request validation.

### Helm (infrastructure layer)

**`helm/values.yaml`**
- Extended the `vllmApiKey` schema annotation from `type:[string, object]` to `type:[string, array, object]`
- Added inline documentation with examples for all three accepted forms:
  ```yaml
  vllmApiKey: "single-key"
  vllmApiKey: "key1,key2,key3"
  vllmApiKey: ["key1", "key2", "key3"]
  vllmApiKey: {secretName: my-secret, secretKey: api-key}
  ```

**`helm/templates/secrets.yaml`**
- Added a `kindIs "slice"` branch: when `vllmApiKey` is a YAML list, entries are joined with commas (`join ","`) before being base64-encoded into the Kubernetes Secret. The Secret always stores a single comma-separated string, the format vLLM natively expects for `--api-key`.

**`helm/templates/deployment-vllm-multi.yaml`** and **`helm/templates/deployment-router.yaml`**
- Extended the `kindIs "string"` guard to `or (kindIs "string") (kindIs "slice")` so that list-typed keys also correctly resolve the `VLLM_API_KEY` env var from the generated Secret.

### Router (application layer)

**`src/vllm_router/auth.py`** *(new)*
- `_parse_api_keys(raw)`: splits a comma-separated string into a `frozenset`, stripping whitespace and ignoring empty segments
- `get_allowed_api_keys()`: reads `VLLM_API_KEY` from the environment at request time
- `verify_api_key(request)`: FastAPI dependency that:
  - Is a **no-op when `VLLM_API_KEY` is unset** (preserves backward compatibility for unauthenticated deployments)
  - Returns `HTTP 401` if the `Authorization` header is missing or not a `Bearer` token
  - Returns `HTTP 401` if the token is not in the allowed set
  - Accepts **any** of the configured keys, enabling true multi-tenant access

**`src/vllm_router/app.py`**
[...]

Issue #655, opened 2025-08-22
feature: Check sleep mode only interacting with the /is_sleeping endpoint


jonoillar:
### Describe the feature

**Scenario**:

I was testing the router with a fake-server, very similar to this one: https://github.com/vllm-project/production-stack/blob/main/src/tests/perftest/fake-openai-server.py

I packaged the app within a docker image:

```
FROM python:3.10-slim AS build

# Install dependencies
COPY requirements.txt .
RUN apt-get update && apt-get install -y gcc
RUN pip install --no-cache-dir --user -r requirements.txt

# Set working directory
WORKDIR /opt/project


# Copy application code
COPY . .
# Set PYTHONPATH
ENV PYTHONPATH=/opt/project

CMD ["python3", "./fake-openai-server.py", "--host", "0.0.0.0", "--port", "8000"]
```

And deploying it within kubernetes, here is the pod spec:

```
apiVersion: v1
kind: Pod
metadata:
  generateName: vllm-k8s-test-fake-server-fake-model-deployment-vllm-5bc7967b99-
  generation: 1
  labels:
    environment: test
    helm-release-name: vllm-k8s-test-fake-server
    model: fake-model
    pod-template-hash: 5bc7967b99
    release: test
  name: vllm-k8s-test-fake-server-fake-model-deployment-vllm-5bc79kndb9
  namespace: default
  ownerReferences:
  - apiVersion: apps/v1
    blockOwnerDeletion: true
    controller: true
    kind: ReplicaSet
    name: vllm-k8s-test-fake-server-fake-model-deployment-vllm-5bc7967b99
spec:
  containers:
  - image: fake-server:latest
    imagePullPolicy: IfNotPresent
    livenessProbe:
      failureThreshold: 3
      httpGet:
        path: /health
        port: 8000
        scheme: HTTP
      initialDelaySeconds: 15
      periodSeconds: 10
      successThreshold: 1
      timeoutSeconds: 1
    name: vllm
    ports:
    - containerPort: 8000
      name: container-port
      protocol: TCP
    - containerPort: 55555
      name: zmq-port
      protocol: TCP
    - containerPort: 9999
      name: ucx-port
      protocol: TCP
    startupProbe:
      failureThreshold: 60
      httpGet:
        path: /health
        port: 8000
        scheme: HTTP
      initialDelaySeconds: 15
      periodSeconds: 10
      successThreshold: 1
      timeoutSeconds: 1
```

[...]

PR #1096, opened 2026-09-18
[Bugfix][Router] Handle container.command=None in the service-name sleep-mode check

chrikrah:
`K8sServiceNameServiceDiscovery._check_engine_sleep_mode` iterates `container.command` directly at `service_discovery.py:1032`. A container that does not override the image entrypoint has `command=None`, which is an ordinary deployment, and the loop raises:

```
TypeError: 'NoneType' object is not iterable
  src/vllm_router/service_discovery.py:1032
```

`except client.rest.ApiException` does not catch a `TypeError`. It leaves `_add_engine` and reaches `_watch_engines`, whose outer `except Exception` logs and sleeps 0.5 seconds, then restarts the watch from a fresh list. Every service on that watch is affected, not only the one with no command.

## The same guard already exists one class up

`K8sPodIPServiceDiscovery._check_engine_sleep_mode` at `:541`:

```python
if (
    not container.command
    or "--enable-sleep-mode" in container.command
):
    enable_sleep_mode = True
break
```

#696 added that for #655 and touched the pod-IP class only. This change applies the same guard, with the same semantics, to the service-name class. No explicit command means sleep mode is assumed on, which is what #696 decided.

## Tests

```
$ uv run pytest src/tests -q
234 passed

$ uv run pre-commit run --files src/vllm_router/service_discovery.py src/tests/test_k8s_service_name_sleep_mode.py
black....................................................................Passed
isort....................................................................Passed
ruff (legacy alias)......................................................Passed
codespell................................................................Passed
```

Three new cases: no command, command carrying the flag, command without it. Revert `service_discovery.py` alone, keep them, and only the first fails, with the `TypeError` above.

The instance is built with `__new__` and given only the attributes the method touches. `test_k8s_pod_ip_service_discovery.py` does the same, because the real `__init__` loads a kubeconfig and starts a watcher thread.

Not run against a live cluster. The `k8s_api` is a mock.

## Scope

#655 is open and asks for something larger: read sleep mode from the `/is_sleeping` endpoint rather than inspect the container args. This change does not do that. It makes the current mechanism behave the same in both discovery classes, and it remains a candidate for deletion if #655 lands.

## Duplicate search

Six open pull requests touch `src/vllm_router/service_discovery.py`: #1092, #1084, #1064, #1004, #945 and #937. None mentions `_check_engine_sleep_mode` or `--enable-sleep-mode`.


chrikrah (commented):
Taken the second finding, not the first.

**The test helper.** Applied in `9aeb371`. The cascade made `service.spec.selector` a `MagicMock` whose `items()` iterated empty, so the label selector came out as an empty string. The three cases reached the `container.command` branch through a selector nobody chose. It now carries `{"app": "vllm"}`. 234 passed.

**The `except` clause.** Leaving it as `client.rest.ApiException`.

Widening it to `Exception` would stop this crash. It would also stop every future one in this method the same way, including the next `AttributeError` that means the code is wrong rather than the cluster unreachable.

That is a decision about whether service discovery should swallow programming errors. It applies equally to `K8sPodIPServiceDiscovery._check_engine_sleep_mode` and to the other `except client.rest.ApiException` handlers in this file, so it belongs in its own change rather than inside a one-line fix for a known cause.

Behind the narrow fix sits the watcher's own outer `except Exception` in `_watch_engines`, which already catches anything this method raises.

Happy to open the wider change separately if you want it.

Shaoting-Feng (approved):
LGTM

PR #950, opened 2026-05-12
[Bugfix] Fix /v1/audio/transcriptions endpoint when not using roundrobin

guba-elleschr:
The /v1/audio/transcriptions endpoint was failing with an error due to missing a `request_json` argument when set to any routing logic besides `roundrobin`.
```TypeError: KvawareRouter.route_request() missing 1 required positional argument: 'request_json'```

This is due to `route_request` requiring different parameters depending on routing logic.

This change copies the logic from `route_general_request` to match the required parameters [here](https://github.com/vllm-project/production-stack/blob/main/src/vllm_router/services/request_service/request.py#L558).

After the change, it successfully routes the traffic:
```
[2026-05-12 13:59:15,723] DEBUG: ==== Enter audio_transcriptions ==== (request.py:1119:vllm_router.services.request_service.request)
[2026-05-12 13:59:15,723] DEBUG: Received upload: example.wav (application/octet-stream) (request.py:1120:vllm_router.services.request_service.request)
[2026-05-12 13:59:15,723] DEBUG: Params: model=qwen3-asr prompt=None response_format='json' temperature=None language=en stream=False (request.py:1121:vllm_router.services.request_service.request)
[2026-05-12 13:59:15,731] DEBUG: Lookup return message: LookupRetMsg(event_id='Lookupf7fe41f7-7b6e-4dfc-a5cd-a5e2ad4dd12c', layout_info={}) (routing_logic.py:382:vllm_router.routers.routing_logic)
[2026-05-12 13:59:15,731] DEBUG: Fallback to using session id: None (routing_logic.py:395:vllm_router.routers.routing_logic)
[2026-05-12 13:59:15,731] INFO: Proxying multi-part form request for model qwen3-asr to http://10.244.6.164:8000 (request.py:1232:vllm_router.services.request_service.request)
INFO:     10.244.6.84:34549 - "POST /v1/audio/transcriptions HTTP/1.1" 200 OK
```

PR #824, opened 2026-02-05
[Feat] Add token redaction for logger debug

shernshiou:
This PR implements a new feature to redact the token by default
If the user set `disable_token_redaction` to True, the redaction will be skipped. 

FIX #819 

**BEFORE SUBMITTING, PLEASE READ THE CHECKLIST BELOW AND FILL IN THE DESCRIPTION ABOVE**

---

- [x] Make sure the code changes pass the [pre-commit](https://github.com/vllm-project/production-stack/blob/main/CONTRIBUTING.md) checks.
- [x] Sign-off your commit by using <code>-s</code> when doing <code>git commit</code>
- [x] Try to classify PRs for easy understanding of the type of changes, such as `[Bugfix]`, `[Feat]`, and `[CI]`.

<details>

<summary><b> Detailed Checklist (Click to Expand) </b></summary>

<p>Thank you for your contribution to production-stack! Before submitting the pull request, please ensure the PR meets the following criteria. This helps us maintain the code quality and improve the efficiency of the review process.</p>

<h3>PR Title and Classification</h3>
<p>Please try to classify PRs for easy understanding of the type of changes. The PR title is prefixed appropriately to indicate the type of change. Please use one of the following:</p>
<ul>
    <li><code>[Bugfix]</code> for bug fixes.</li>
    <li><code>[CI/Build]</code> for build or continuous integration improvements.</li>
    <li><code>[Doc]</code> for documentation fixes and improvements.</li>
    <li><code>[Feat]</code> for new features in the cluster (e.g., autoscaling, disaggregated prefill, etc.).</li>
    <li><code>[Router]</code> for changes to the <code>vllm_router</code> (e.g., routing algorithm, router observability, etc.).</li>
    <li><code>[Misc]</code> for PRs that do not fit the above categories. Please use this sparingly.</li>
</ul>
<p><strong>Note:</strong> If the PR spans more than one category, please include all relevant prefixes.</p>

<h3>Code Quality</h3>

<p>The PR need to meet the following code quality standards:</p>

<ul>
    <li>Pass all linter checks. Please use <code>pre-commit</code> to format your code. See <code>README.md</code> for installation.</li>
    <li>The code need to be well-documented to ensure future contributors can easily understand the code.</li>
    <li> Please include sufficient tests to ensure the change is stay correct and robust. This includes both unit tests and integration tests.</li>
</ul>

<h3>DCO and Signed-off-by</h3>
[...]

PR #1097, opened 2026-09-18
[Bugfix][Router] Log request headers as a lazy argument so they can be redacted

chrikrah:
`TokenRedactionFilter` only inspects `record.args` for a Starlette `Headers` or `MutableHeaders` object, at `src/vllm_router/log.py:163`.

The session-extraction line formats the headers into the message instead, at `src/vllm_router/services/request_service/request.py:610`:

```python
logger.debug(f"Debug session extraction - Request headers: {dict(request.headers)}")
```

`record.args` is then empty. The filter has nothing to act on, so `Authorization` and `Cookie` print in full at DEBUG level for every request through `route_general_request`.

Measured with the real filter, both forms of the same call:

```
CURRENT : Debug session extraction - Request headers: {'authorization': 'Bearer super-secret-token', 'host': 'x'}
LAZY    : Debug session extraction - Request headers: {'authorization': 'Bearer ****', 'host': 'x'}
```

## Second change, same class

`init_logger` attaches `TokenRedactionFilter` to the stdout handler only, at `log.py:207`. Each handler owns its filter chain, and the stdout handler also carries `MaxLevelFilter(logging.INFO)`, so nothing redacts a record at WARNING or above.

No call site logs headers at that level today. This closes the hole rather than a live leak, and it costs one line.

## Tests

```
$ uv run pytest src/tests -q
235 passed

$ uv run pre-commit run --files src/vllm_router/log.py src/vllm_router/services/request_service/request.py src/tests/test_log.py src/tests/test_request_header_redaction.py
black....................................................................Passed
isort....................................................................Passed
ruff (legacy alias)......................................................Passed
codespell................................................................Passed
```

Three new cases. One pins that a pre-formatted dict is not redacted while a lazy argument is, which is the rule the call site has to follow. One reads the call site's own source, so a later edit back to an f-string fails rather than silently reopening this. Two cover the stderr handler.

Revert both production files, keep the tests, and the two files give `3 failed, 24 passed`.

## Scope

The message reads identically once redaction has run, so no log format changed for anyone whose headers carry nothing sensitive. `_SENSITIVE_HEADERS` and `_redact_value` are untouched.

## Duplicate search

Eleven open pull requests touch `src/vllm_router/services/request_service/request.py`: #1095, #1082, #1072, #1067, #1040, #1023, #1020, #1001, #951, #950 and #939. Not one mentions `Request headers`.

No open pull request touches `src/vllm_router/log.py`. `gh pr list --state all --search "TokenRedactionFilter"` returns only #824, the merged change that added the filter.


chrikrah (commented):
Both findings are right. One is applied, one is recorded rather than acted on.

**The source-matching test.** Applied in `aff7554`. A string match breaks on quote style and on a formatter rewrapping the line, exactly as you say.

A `caplog` test loses the point, though. Logging with a lazy argument works whatever the call site does, so such a test passes with the production change reverted and pins nothing.

The test now parses the function, finds the `logger.debug` call whose first argument mentions `Request headers`, and asserts it has a second argument. That survives reformatting and requoting. Reverting `request.py` alone still gives `1 failed, 3 passed`.

**The format inconsistency.** Real, and deliberate upstream of this change.

`TokenRedactionFilter` replaces the `Headers` object only when it redacted something, so this line reads as a dict with a secret present and as `Headers({...})` without one. `test_filter_preserves_non_sensitive_headers` asserts that today:

```python
assert isinstance(log_record.args[0], Headers)
```

Always converting would change the filter's contract and that test, which is a separate decision from this one. I added a case pinning the current behaviour and naming the consequence, so whoever revisits it sees both halves.

Happy to make the filter always convert, in its own change, if that is what you want.

